PR #12, opened 2020-07-28
Tari Script

CjS77:
Tari script is heavily inspired by Bitcoin script in that:
* It is non-Turing complete
* Is stack based
* Shares many of the same opcodes

For script safety, the Rust type system is leveraged so that
* Opcode operations on inappropriate input data is impossible
* Bounds checking is done on arithmetic operations

Tari script differs from Bitcoin script in that:
* Succesful execution requires exactly zero to be on the stack after
completion
* Blake256 and Curve25519 are the default hash and elliptic curves
respectively.

For a PoC, pay-to-public-key-hash (P2PKH) is implemented in the tests.

Serde support is provided.

## OpCodeS:

```
#[derive(Debug, Clone, PartialEq, Eq)]
pub enum Opcode {
    /// Push the associated 32-byte value onto the stack
    PushHash(Box<HashValue>),
    /// Hash to top stack element with the Blake256 hash function and push the result to the stack
    HashBlake256,
    /// Fail the script immediately. (Must be executed.)
    Return,
    /// Drops the top stack item
    Drop,
    /// Duplicates the top stack item
    Dup,
    /// Reverse rotation. The top stack item moves into 3rd place, abc => bca
    RevRot,
    /// Pop two items and push their sum
    Add,
    /// Pop two items and push the second minus the top
    Sub,
    /// Pop the public key and then the signature. If the signature signs the script, push 0 to the stack, otherwise
    /// push 1
    CheckSig,
    /// As for CheckSig, but aborts immediately if the signature is invalid. As opposed to Bitcoin, it pushes a zero
    /// to the stack if successful
    CheckSigVerify,
    /// Pushes 0, if the inputs are exactly equal, 1 otherwise
    Equal,
    /// Pushes 0, if the inputs are exactly equal, aborts otherwise
    EqualVerify,
}
```

## Stack types

```
pub enum StackItem {
pub enum StackItem {
    Number(i64),
    Hash(HashValue),
    Commitment(PedersenCommitment),
    PublicKey(RistrettoPublicKey),
    Signature(RistrettoSchnorr),
}
```

The tests demonstrate how scripts can be constructed and executed:

```
    fn pay_to_public_key_hash() {
        let k = RistrettoSecretKey::from_hex("7212ac93ee205cdbbb57c4f0f815fbf8db25b4d04d3532e2262e31907d82c700").unwrap();
        let p = RistrettoPublicKey::from_secret_key(&k); // 56c0fa32558d6edc0916baa26b48e745de834571534ca253ea82435f08ebbc7c
        let r =
            RistrettoSecretKey::from_hex("193ee873f3de511eda8ae387db6498f3d194d31a130a94cdf13dc5890ec1ad0f").unwrap();
        let hash = Blake256::digest(p.as_bytes());
        let pkh = to_hash(hash.as_slice()); // ae2337ce44f9ebb6169c863ec168046cb35ab4ef7aa9ed4f5f1f669bb74b09e5
        let script = script!(Dup HashBlake256 PushHash(pkh) EqualVerify Drop CheckSig);
        let hex_script = "71b07aae2337ce44f9ebb6169c863ec168046cb35ab4ef7aa9ed4f5f1f669bb74b09e58170ac";

        let m = script.script_message(&p).unwrap();
        let sig = RistrettoSchnorr::sign(k, r, m.as_bytes()).unwrap();
        // The top of the stack is the right-most element!
        let inputs = inputs!(sig, p);
        assert_eq!(script.execute(&inputs), Ok(()));
    }
```

### DIFF
--- a/src/ristretto/utils.rs
+++ b/src/ristretto/utils.rs
@@ -1,17 +1,12 @@
-// Copyright 2019 The Tari Project
-//
+// Copyright 2020. The Tari Project
 // Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
 // following conditions are met:
-//
 // 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
 // disclaimer.
-//
 // 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
 // following disclaimer in the documentation and/or other materials provided with the distribution.
-//
 // 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
 // products derived from this software without specific prior written permission.
-//
 // THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND ANY EXPRESS OR IMPLIED WARRANTIES,
 // INCLUDING, BUT NOT LIMITED TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
 // DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
@@ -20,21 +15,43 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
-pub mod constants;
-pub mod dalek_range_proof;
-pub mod musig;
-pub mod pedersen;
-pub mod ristretto_keys;
-pub mod ristretto_sig;
-pub mod serialize;
-pub mod utils;
+//! Handy utility functions for use in tests and demo scripts
 
-// Re-export
-pub use self::{
-    ristretto_keys::{RistrettoPublicKey, RistrettoSecretKey},
-    ristretto_sig::RistrettoSchnorr,
+use crate::{
+    keys::PublicKey,
+    ristretto::{RistrettoPublicKey, RistrettoSchnorr, RistrettoSecretKey},
+    signatures::SchnorrSignatureError,
 };
+use digest::Digest;
+use tari_utilities::ByteArray;
 
-// test modules
-#[cfg(test)]
-mod test_common;
+/// A set of keys and it's associated signature
+pub struct SignatureSet {
+    pub nonce: RistrettoSecretKey,
+    pub public_nonce: RistrettoPublicKey,
+    pub message: Vec<u8>,
+    pub signature: RistrettoSchnorr,
+}
+
+/// Generate a random keypair and a signature for the provided message
+///
+/// # Panics
+///
+/// The function panics if it cannot generate a suitable signature
+pub fn sign<D: Digest>(
+    private_key: &RistrettoSecretKey,
+    message: &[u8],
+) -> Result<SignatureSet, SchnorrSignatureError>
+{
+    let mut rng = rand::thread_rng();
+    let (nonce, public_nonce) = RistrettoPublicKey::random_keypair(&mut rng);
+    let message = D::new().chain(public_nonce.as_bytes()).chain(message).result().to_vec();
+    let e = RistrettoSecretKey::from_bytes(&message).map_err(|_| SchnorrSignatureError::InvalidChallenge)?;
+    let s = RistrettoSchnorr::sign(private_key.clone(), nonce.clone(), e.as_bytes())?;
+    Ok(SignatureSet {
+        nonce,
+        public_nonce,
+        message,
+        signature: s,
+    })
+}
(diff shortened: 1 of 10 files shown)


src/script/tari_script.rs
```diff
@@ -0,0 +1,345 @@
...
+            .chain(&self.as_bytes())
+            .result()
+            .to_vec();
+        RistrettoSecretKey::from_bytes(&b).map_err(|_| ScriptError::InvalidSignature)
+    }
+
+    fn execute_opcode(&self, opcode: &Opcode, stack: &mut ExecutionStack) -> Result<(), ScriptError> {
+        use StackItem::*;
+        match opcode {
+            Opcode::Return => Err(ScriptError::Return),
```
SWvheerden, line 151: Why is the return opcode seen as an error?

CjS77: That's how it's defined in Bitcoin. OP_RETURN always fails

CjS77: AFAIK any Bitcoin UTXOs with OP_RETURN in their scripts are burnt

CjS77:
I think I can make the input macro more ergonomic.

From
`let inputs = inputs!(PublicKey(p1), PublicKey(p2), Signature(s1), Signature(s2));`

to

`let inputs = inputs!(p1 p2 s1 s2);`

By implementing `From T for StackItem`

src/script/tari_script.rs
```diff
@@ -0,0 +1,345 @@
...
+macro_rules! script {
+    ($($opcode:ident) +) => {{
+        use crate::script::TariScript;
+        use crate::script::Opcode;
+        let script = vec![$(Opcode::$opcode),+];
+        TariScript::new(script)
+    }}
+}
+
+macro_rules! inputs {
```
sdbondi, line 39: Missing a `#[macro_export]`?

src/script/tari_script.rs
```diff
@@ -0,0 +1,345 @@
...
+
+    /// Calculate the message hash that CHECKSIG uses to verify signatures
+    pub fn script_message(&self, pub_key: &RistrettoPublicKey) -> Result<RistrettoSecretKey, ScriptError> {
+        let b = Blake256::new()
+            .chain(pub_key.as_bytes())
+            .chain(&self.as_bytes())
+            .result()
+            .to_vec();
+        RistrettoSecretKey::from_bytes(&b).map_err(|_| ScriptError::InvalidSignature)
+    }
```
sdbondi, line 146: 90% sure you can save a `to_vec` (clone) here
```suggestion
    pub fn script_message(&self, pub_key: &RistrettoPublicKey) -> Result<RistrettoSecretKey, ScriptError> {
        let b = Blake256::new()
            .chain(pub_key.as_bytes())
            .chain(&self.as_bytes())
            .result();
        RistrettoSecretKey::from_bytes(&b).map_err(|_| ScriptError::InvalidSignature)
    }
```

src/script/tari_script.rs
CjS77: yes

src/script/tari_script.rs
CjS77: That did not work. I also thought `GenericArray` was referencable as a `&[u8]`

src/script/tari_script.rs
```diff
@@ -0,0 +1,345 @@
...
+        match (pk, sig) {
+            (PublicKey(p), Signature(s)) => {
+                let m = self.script_message(&p)?;
+                Ok(s.verify(&p, &m))
+            },
+            (..) => Err(ScriptError::IncompatibleTypes),
+        }
+    }
+
+    fn is_successful(stack: &ExecutionStack) -> Result<(), ScriptError> {
```
mikethetike, line 203: Maybe `ensure_successful` or `assert_successful` is a better name? `is_successful` sounds like it returns a true/false

src/script/tari_script.rs
```diff
@@ -0,0 +1,345 @@
...
+        if stack.size() != 1 {
```
mikethetike, line 204: It might also be helpful in debugging to throw an EmptyStack error here if there's nothing

mikethetike:
> I think I can make the input macro more ergonomic.
> 
> From
> `let inputs = inputs!(PublicKey(p1), PublicKey(p2), Signature(s1), Signature(s2));`
> 
> to
> 
> `let inputs = inputs!(p1 p2 s1 s2);`
> 
> By implementing `From T for StackItem`

I think the commas are nicer if you need to have a line break in them

CjS77:
> > I think I can make the input macro more ergonomic.
> > From
> > `let inputs = inputs!(PublicKey(p1), PublicKey(p2), Signature(s1), Signature(s2));`
> > to
> > `let inputs = inputs!(p1 p2 s1 s2);`
> > By implementing `From T for StackItem`
> 
> I think the commas are nicer if you need to have a line break in them

Maybe, thought maybe keeping it consistent with the script macro was better, but I'm not married to either tbh

src/script/tari_script.rs
CjS77: `.as_slice` did the trick

src/script/tari_script.rs
CjS77: went `stack_is_zero`

src/script/tari_script.rs
CjS77: Maybe we could return the stack height on error, but I don't see the value in a separate error type

src/ristretto/utils.rs
```diff
@@ -0,0 +1,57 @@
...
+use crate::{
+    keys::PublicKey,
+    ristretto::{RistrettoPublicKey, RistrettoSchnorr, RistrettoSecretKey},
+    signatures::SchnorrSignatureError,
+};
+use digest::Digest;
+use tari_utilities::ByteArray;
+
+/// A set of keys and it's associated signature
+pub struct SignatureSet {
```
mikethetike, line 29: nit: If these fields are encapsulated with readonly access, then it means that code can't alter part of the struct.

PR #17, opened 2020-08-21
Add Sha3 hash function

SWvheerden:
Add sha3 as a hash function for tari_crypto
Refactor crypto to better accommodate more than one hash function. 

This Pr adds in the sh3 as a hash function. Its added in the same way as as blake2. This will allow tari to swop hash functions between the two with just the changed of a type. 

The tari_crypto crate is also modified to allow more than one hash function to be supported easily. 
Added a pub use to not break the moving of the blake2 hash function.


mikethetike (approved):
LGTM

mikethetike (approved):

PR #18, opened 2020-08-24
Extra hash functions

SWvheerden:
As per IRC discussion, I have added Blake3 and K12 as experimental proof of concept hash functions to tari_crypto. 
They both implement the same traits to enable swapping between Sha3, Blake2, Blake3 and K12 very easy. 

This PR expands on #17 and uses the same refactoring for multiple hash functions. This will allow tari_crypto to use these hash functions as well as experiment with them in tari_core.


src/hash/k12.rs
```diff
@@ -0,0 +1,92 @@
...
+    FixedOutput,
+    Input,
+    VariableOutput,
+};
+use k12::{
+    digest::{ExtendableOutput, Reset, Update},
+    KangarooTwelve,
+};
+
+/// A convenience wrapper produce 256 bit hashes from Blake2b
```
mikethetike, line 34: Doc out of date?

src/hash/blake3.rs
```diff
@@ -0,0 +1,93 @@
...
+
+use blake3::Hasher;
+use digest::{
+    generic_array::{typenum::U32, GenericArray},
+    FixedOutput,
+    Input,
+    Reset,
+};
+
+/// A convenience wrapper produce 256 bit hashes from Blake2b
```
mikethetike, line 31: Doc out of date

SWvheerden: thanks

src/hash/k12.rs
SWvheerden: thanks

mikethetike (approved):

PR #24, opened 2020-11-02
Add signing methods for pre-determined nonces

mikethetike:
In some cases, the public nonce used in the signature is committed and
used in the challenge. These extra methods allow signing with a private
nonce to accommodate that. There is also support for signing a pre-hashed
byte array instead of a string message

### DIFF
--- a/src/wasm/key_utils.rs
+++ b/src/wasm/key_utils.rs
@@ -32,7 +32,7 @@ use crate::{
 use blake2::Digest;
 use rand::rngs::OsRng;
 use serde::{Deserialize, Serialize};
-use tari_utilities::hex::Hex;
+use tari_utilities::hex::{from_hex, Hex};
 use wasm_bindgen::prelude::*;
 
 #[derive(Serialize, Deserialize)]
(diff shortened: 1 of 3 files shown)


src/wasm/key_utils.rs
```diff
@@ -87,15 +87,62 @@ pub fn sign(private_key: &str, msg: &str) -> JsValue {
...
 }
 
-#[allow(non_snake_case)]
-pub(crate) fn sign_with_key(k: &RistrettoSecretKey, msg: &str, result: &mut SignResult) {
-    let (r, R) = RistrettoPublicKey::random_keypair(&mut OsRng);
+/// Generate a Schnorr signature of a challenge that has already been hashed using the given private
+/// key and a specified private nonce. DO NOT reuse nonces. This method is provide for cases where a
+/// public nonce has been used
+/// in the message.
+#[wasm_bindgen]
```
CjS77, line 98: I think you meant pre-hashed message, but can you explain what "hashed using the given private key" means? You usually hash with a hashing algorithm.

CjS77 (approved):
Looks ok, but one docstring typo.

src/wasm/key_utils.rs
mikethetike: Yeah needs a comma or something. PHRASING.

```
Generate a Schnorr signature of a challenge that has already been hashed, using the given private
```

CjS77 (approved):

PR #28, opened 2020-12-21
pubkey_from_hex WASM function

sdbondi:
pubkey_from_hex function deserializes tari public key hex (Ristretto)
and returns a javascript object with the decoded public key.
If the public key does not represent a valid public key, boolean `false`
is returned. This is useful for validating the public key in javascript
environments.


CjS77 (approved):
LGTM

PR #31, opened 2020-12-29
Fix grcov action

CjS77:
* Only run lib tests (doc tests shouldn't count towards coverage)
* Fix stack overflow issue -- don't run FFI binding generation and change panic strategy


mikethetike (approved):
Neat

PR #32, opened 2021-02-26
Add compile flags to dependencies for wasm32 target

dunnock:
When compiling tari_crypto as dependency, when we do not need to include it's exports it is failing now with compile error:
```
  process didn't exit successfully:  (exit code: 1)
  --- stdout
  TARGET = Some(wasm32-unknown-unknown)
```

This change fixes that issue. Worked for me on when I updated dependency to:
```
tari_crypto = { version = "0.8.0", git = "https://github.com/dunnock/tari-crypto", branch = "dunnock-wasm32-dependencies" }
```


dunnock:
Closing in favour of https://github.com/tari-project/tari-crypto/pull/33

PR #33, opened 2021-02-26
Expose fine-grained control of cc feature flags

CjS77:
This PR resolves two inter-related issues:

* Tari_crypto WASM doesn't work on Safari currently
* wasm_pack won't compile `clear_on_drop` (need to use either `no_cc` or `nightly` flags)

The latter was fixed by including the `no-cc` flag in the `wasm` flag. This causes issues with libraries that use tari_crypto as a dependency for building wasm, but don't want to export the wasm functions defined here. So they either cannot build, or they switch on the `wasm` feature, but then Safari isn't supported.

This PR simply decouples the two issues.

You can export wasm with the `wasm` flag; and control the `clear_on_drop` compilation using either the `no_cc` or `no_cc_nightly` flags.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ categories = ["cryptography"]
 homepage = "https://tari.com"
 readme = "README.md"
 license = "BSD-3-Clause"
-version = "0.8.2"
+version = "0.8.3"
 edition = "2018"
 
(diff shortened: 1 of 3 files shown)


mikethetike (approved):
LGTM

PR #36, opened 2021-03-02
Update TariScript/opcodes and implementation to latest RFC-0202 

delta1:
Opcodes and script logic updated from [RFC-0202](https://rfc.tari.com/RFC-0202_TariScriptOpcodes.html). Includes tests for new opcodes.

### DIFF
--- a/src/script/op_codes.rs
+++ b/src/script/op_codes.rs
@@ -15,114 +15,339 @@
 // WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
+use crate::ristretto::RistrettoPublicKey;
 use std::{fmt, ops::Deref};
-use tari_utilities::hex::Hex;
+use tari_utilities::{hex::Hex, ByteArray};
+
+use super::ScriptError;
 
 pub type HashValue = [u8; 32];
+pub type Message = [u8; 32];
 
 /// Convert a slice into a HashValue.
 ///
 /// # Panics
 ///
 /// The function does not check slice for length at all.  You need to check this / guarantee it yourself.
-pub fn to_hash(slice: &[u8]) -> HashValue {
+pub fn slice_to_hash(slice: &[u8]) -> HashValue {
     let mut hash = [0u8; 32];
     hash.copy_from_slice(slice);
     hash
 }
 
 /// Convert a slice into a Boxed HashValue
-pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
-    Box::new(to_hash(slice))
+pub fn slice_to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
+    Box::new(slice_to_hash(slice))
 }
 
-// Opcode constants: Contextual
-pub const OP_PUSH_CHAIN_HEIGHT: u8 = 0x50;
+/// Convert a slice into a Message.
+///
+/// # Panics
+///
+/// The function does not check slice for length at all.  You need to check this / guarantee it yourself.
+pub fn slice_to_message(slice: &[u8]) -> Message {
+    let mut msg = [0u8; 32];
+    msg.copy_from_slice(slice);
+    msg
+}
 
-// Opcode constants: Script termination
-pub const OP_RETURN: u8 = 0x60;
+/// Convert a slice into a Boxed Message
+pub fn slice_to_boxed_message(slice: &[u8]) -> Box<Message> {
+    Box::new(slice_to_message(slice))
+}
 
-// Opcode constants: Stack manipulation
+/// Convert a slice of little endian bytes into a u64.
+///
+/// # Panics
+///
+/// The function does not check slice for length at all.  You need to check this / guarantee it yourself.
+fn slice_to_u64(slice: &[u8]) -> u64 {
+    let mut num = [0u8; 8];
+    num.copy_from_slice(slice);
+    u64::from_le_bytes(num)
+}
+
+/// Convert a slice of little endian bytes into an i64.
+///
+/// # Panics
+///
+/// The function does not check slice for length at all.  You need to check this / guarantee it yourself.
+fn slice_to_i64(slice: &[u8]) -> i64 {
+    let mut num = [0u8; 8];
+    num.copy_from_slice(slice);
+    i64::from_le_bytes(num)
+}
+
+// Opcode constants: Block Height Checks
+pub const OP_CHECK_HEIGHT_VERIFY: u8 = 0x66;
+pub const OP_CHECK_HEIGHT: u8 = 0x67;
+pub const OP_COMPARE_HEIGHT_VERIFY: u8 = 0x68;
+pub const OP_COMPARE_HEIGHT: u8 = 0x69;
+
+// Opcode constants: Stack Manipulation
 pub const OP_DROP: u8 = 0x70;
 pub const OP_DUP: u8 = 0x71;
 pub const OP_REV_ROT: u8 = 0x72;
 pub const OP_PUSH_HASH: u8 = 0x7a;
 pub const OP_PUSH_ZERO: u8 = 0x7b;
+pub const OP_NOP: u8 = 0x73;
+pub const OP_PUSH_ONE: u8 = 0x7c;
+pub const OP_PUSH_INT: u8 = 0x7d;
+pub const OP_PUSH_PUBKEY: u8 = 0x7e;
 
-// Opcode constants: Comparisons
+// Opcode constants: Math Operations
 pub const OP_EQUAL: u8 = 0x80;
 pub const OP_EQUAL_VERIFY: u8 = 0x81;
-
-// Opcode constants: Arithmetic
 pub const OP_ADD: u8 = 0x93;
 pub const OP_SUB: u8 = 0x94;
+pub const OP_GE_ZERO: u8 = 0x82;
+pub const OP_GT_ZERO: u8 = 0x83;
+pub const OP_LE_ZERO: u8 = 0x84;
+pub const OP_LT_ZERO: u8 = 0x85;
 
-// Opcode constants: Cryptography
+// Opcode constants: Boolean Logic
+pub const OP_OR_VERIFY: u8 = 0x64;
+pub const OP_OR: u8 = 0x65;
+
+// Opcode constants: Cryptographic Operations
 pub const OP_CHECK_SIG: u8 = 0xac;
 pub const OP_CHECK_SIG_VERIFY: u8 = 0xad;
 pub const OP_HASH_BLAKE256: u8 = 0xb0;
+pub const OP_HASH_SHA256: u8 = 0xb1;
+pub const OP_HASH_SHA3: u8 = 0xb2;
+
+// Opcode constants: Miscellaneous
+pub const OP_RETURN: u8 = 0x60;
+pub const OP_IF_THEN: u8 = 0x61;
+pub const OP_ELSE: u8 = 0x62;
+pub const OP_END_IF: u8 = 0x63;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub enum Opcode {
-    /// Push the current chain height onto the stack
-    PushHeight,
-    /// Push the associated 32-byte value onto the stack
-    PushHash(Box<HashValue>),
-    /// Push a zero onto the stack
+    // Block Height Checks
+    /// Compare the current block height to height. Fails with VERIFY_FAILED if the block height < height.
+    CheckHeightVerify(u64),
+    /// Pushes the value of (the current tip height - height) to the stack. In other words, the top of the stack will
+    /// hold the height difference between height and the current height. If the chain has progressed beyond
+    /// height, the value is positive; and negative if the chain has yet to reach height. Fails with STACK_OVERFLOW
+    /// if the stack would exceed the max stack height.
+    CheckHeight(u64),
+    /// Pops the top of the stack as height and compares it to the current block height. Fails with INVALID_INPUT
+    /// if there is not a valid integer value on top of the stack. Fails with EMPTY_STACK if the stack is empty.
+    /// Fails with VERIFY_FAILED if the block height < height.
+    CompareHeightVerify,
+    /// Pops the top of the stack as height, then pushes the value of (height - the current height) to the stack. In
+    /// other words, this opcode replaces the top of the stack with the difference between that value and the current
+    /// height. Fails with INVALID_INPUT if there is not a valid integer value on top of the stack. Fails with
+    /// EMPTY_STACK if the stack is empty.
+    CompareHeight,
+
+    // Stack Manipulation
+    /// No op. Does nothing. Never fails.
+    Nop,
+    /// Pushes a zero onto the stack. This is a very common opcode and has the same effect as PushInt(0) but is more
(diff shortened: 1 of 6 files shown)


CjS77 (approved):
LGTM so far

delta1:
@philipr-za ^

src/script/tari_script.rs
```diff
@@ -138,41 +196,193 @@ impl TariScript {
...
+
+    fn handle_ordering(stack: &mut ExecutionStack, valid_orderings: Vec<Ordering>) -> Result<(), ScriptError> {
+        let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
+        let stack_number = if let StackItem::Number(n) = item {
+            Ok(n)
+        } else {
+            Err(ScriptError::InvalidInput)
+        }?;
+
+        if valid_orderings.iter().any(|o| &stack_number.cmp(&0) == o) {
```
sdbondi, line 312: Only need to do the `cmp` once

```suggestion
        let cmp = stack_number.cmp(0);
        if valid_orderings.contains(&cmp) {
```

sdbondi (commented):
Looks great - few fairly minor comments

delta1:
Thanks for the feedback @sdbondi! Will take a look and revert.

src/script/stack.rs
```diff
@@ -188,6 +188,21 @@ impl ExecutionStack {
...
+        let mut items = Vec::new();
+
+        for _ in 0..num_items {
+            match self.items.pop() {
+                Some(item) => items.insert(0, item),
+                None => return Err(ScriptError::StackUnderflow),
+            }
+        }
+
+        Ok(items)
```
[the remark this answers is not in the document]
delta1: Great catch

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
 pub enum Opcode {
-    /// Push the current chain height onto the stack
-    PushHeight,
-    /// Push the associated 32-byte value onto the stack
-    PushHash(Box<HashValue>),
-    /// Push a zero onto the stack
+    // Block Height Checks
+    /// Compare the current block height to height. Fails with VERIFY_FAILED if the block height < height.
+    CheckHeightVerify(u64),
+    /// Pushes the value of (the current height - height) to the stack. In other words, the top of the stack will hold
```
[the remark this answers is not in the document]
delta1: I don't have a strong opinion on this one

src/script/tari_script.rs
```diff
@@ -47,26 +52,76 @@ macro_rules! script {
...
+            // reset the script state
+            self.state = ScriptState::default();
+
+            stack.pop().ok_or(ScriptError::NonUnitLengthStack)
+        } else {
+            Err(ScriptError::NonUnitLengthStack)
+        }
+    }
+
+    fn should_execute(&mut self, opcode: &Opcode) -> Result<bool, ScriptError> {
```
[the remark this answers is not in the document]
delta1: Will do

src/script/tari_script.rs
```diff
@@ -138,41 +196,193 @@ impl TariScript {
...
+        stack.push(item)
+    }
+
+    fn handle_compare_height_verify(stack: &mut ExecutionStack, block_height: u64) -> Result<(), ScriptError> {
+        let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
+        let target_height = if let StackItem::Number(n) = item {
+            Ok(n)
+        } else {
+            Err(ScriptError::InvalidInput)
+        }?;
```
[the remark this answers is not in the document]
delta1: Not sure I prefer this, but too bothered I'll update it. Does it pass clippy?

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
+    Box::new(to_message(slice))
+}
 
-// Opcode constants: Stack manipulation
+/// Convert a slice of little endian bytes into a u64.
+///
+/// # Panics
+///
+/// The function does not check slice for length at all.  You need to check this / guarantee it yourself.
+pub fn to_u64(slice: &[u8]) -> u64 {
```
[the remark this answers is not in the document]
delta1: Good point

src/script/tari_script.rs
```diff
@@ -138,41 +196,193 @@ impl TariScript {
...
+            Nop => Ok(()),
+            PushZero => stack.push(Number(0)),
+            PushOne => stack.push(Number(1)),
+            PushHash(h) => stack.push(Hash(*h.clone())),
+            PushInt(n) => stack.push(Number(*n)),
+            PushPubKey(p) => stack.push(PublicKey(*p.clone())),
+            Drop => TariScript::handle_drop(stack),
+            Dup => TariScript::handle_dup(stack),
+            RevRot => stack.push_down(2),
+            GeZero => TariScript::handle_ordering(stack, vec![Ordering::Greater, Ordering::Equal]),
```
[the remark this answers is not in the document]
delta1: Oh very nice, will do

src/script/tari_script.rs
delta1: Nice I'll try that

src/script/tari_script.rs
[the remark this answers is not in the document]
sdbondi: I don't think clippy complains about the `if {Ok()} else{Err()}?`. Not a big deal at all, and I know I'm nitting. It's easy to miss the `?` at the end of the `if` when reading.  Generally not fond of constructing a `Result` just so that `?` can be used when a `match` is clearer.

```rust
// De-sugared `?`
let target_height = match (if let StackItem::Number(n) = item {
            Ok(n)
        } else {
            Err(ScriptError::InvalidInput)
        }) {
Ok(n) => n,
Err(err) => {return Err(err.into()); }
};
```

src/script/op_codes.rs
SWvheerden, line 128: I also don't have a strong opinion on this one, but if its typically called `depth`, we would probably benefit to changing this
But on the comments, I think we should change the one height to tip to make it less ambiguous 
```suggestion
    /// Pushes the value of (the current tip height - height) to the stack. In other words, the top of the stack will hold
```
.

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
+    /// other words, this opcode replaces the top of the stack with the difference between that value and the current
+    /// height. Fails with INVALID_INPUT if there is not a valid integer value on top of the stack. Fails with
+    /// EMPTY_STACK if the stack is empty.
+    CompareHeight,
+
+    // Stack Manipulation
+    /// No op. Does nothing. Never fails.
+    Nop,
+    /// Pushes a zero onto the stack. This is a very common opcode and has the same effect as PushInt(0) but is more
+    /// compact. PushZero can also be interpreted as PushFalse (although no such opcode exists). Fails with
```
SWvheerden, line 147:
```suggestion
    /// compact. Fails with
```
I think we can remove this, as this is part of the previous idea where 1 is script = true, and 0 is script = fail.

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
+    /// STACK_OVERFLOW if the stack would exceed the max stack height.
     PushZero,
-    /// Hash to top stack element with the Blake256 hash function and push the result to the stack
-    HashBlake256,
-    /// Fail the script immediately. (Must be executed.)
-    Return,
-    /// Drops the top stack item
+    /// Pushes a one onto the stack. This is a very common opcode and has the same effect as PushInt(1) but is more
+    /// compact. PushOne can also be interpreted as PushTrue, although no such opcode exists. Fails with STACK_OVERFLOW
```
SWvheerden, line 151:
```suggestion
    /// compact. Fails with STACK_OVERFLOW
```

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
+    /// would exceed the max stack height.
+    PushPubKey(Box<RistrettoPublicKey>),
+    /// Drops the top stack item. Fails with EMPTY_STACK if the stack is empty.
     Drop,
-    /// Duplicates the top stack item
+    /// Duplicates the top stack item. Fails with EMPTY_STACK if the stack is empty. Fails with STACK_OVERFLOW if the
+    /// stack would exceed the max stack height.
     Dup,
-    /// Reverse rotation. The top stack item moves into 3rd place, abc => bca
+    /// Reverse rotation. The top stack item moves into 3rd place, e.g. abc => bca. Fails with EMPTY_STACK if the stack
```
SWvheerden, line 169:
```suggestion
    /// Reverse rotation. The top stack item moves into last place, e.g. abc => bca. Fails with EMPTY_STACK if the stack
```

SWvheerden (dismissed):
LGTM

src/script/op_codes.rs
delta1: Good point will do

src/script/op_codes.rs
delta1: Not sure I agree with this one, and I did take the text directly from the RFC

src/script/tari_script.rs
```diff
@@ -138,41 +196,195 @@ impl TariScript {
...
+        stack.push(item)
+    }
+
+    fn handle_compare_height_verify(stack: &mut ExecutionStack, block_height: u64) -> Result<(), ScriptError> {
+        let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
+        let target_height = if let StackItem::Number(n) = item {
+            Ok(n)
+        } else {
+            Err(ScriptError::InvalidInput)
+        }?;
```
CjS77, line 280:
```
        let stack_number = if let StackItem::Number(n) = item {
            n
        } else {
            return Err(ScriptError::InvalidInput)
        };
```

though I prefer a match here, since it's more explicit and easy to extend if other types get supported in future

```rust
let stack_number = match item {
            StackItem::Number(n) => n,
            _ => return Err(ScriptError::InvalidInput)
        };
```

src/script/tari_script.rs
```diff
@@ -138,41 +196,195 @@ impl TariScript {
...
+            stack.push(StackItem::Number(0))
+        }
+    }
+
+    fn handle_or_verify(stack: &mut ExecutionStack, n: u8) -> Result<(), ScriptError> {
+        let items = stack.pop_num_items(n)?;
+        let item = stack.pop().ok_or(ScriptError::StackUnderflow)?;
+
+        if items.contains(&item) {
+            Ok(())
```
CjS77, line 337: This is close, but still probably better to DRY up with the non_verify version. 

Ditto for all xxx_verify

src/script/tari_script.rs
```diff
@@ -309,50 +500,310 @@ impl fmt::Display for TariScript {
...
+
+        let mut script = script!(OrVerify(2));
+
+        let inputs = inputs!(1, 2, 2, 3);
+        let result = script.execute(&inputs).unwrap();
+        assert_eq!(result, Number(1));
+
+        let inputs = inputs!(1, 2, 3, 4);
+        let err = script.execute(&inputs).unwrap_err();
+        assert!(matches!(err, ScriptError::VerifyFailed));
```
CjS77, line 596: Try some cases with mixed types, Like a 

Stack:
```
pubkey(Pa)
Number(5)
Pubkey(Pa)
```

`Or(2)` executes, as it does for 

```
pubkey(Pa)
Number(5)
Number(5)
```

CjS77 (changes requested):
One or two missed things that block merging. 

And I'd strongly prefer the mutable state to go away

src/script/op_codes.rs
```diff
@@ -36,92 +38,301 @@ pub fn to_boxed_hash(slice: &[u8]) -> Box<HashValue> {
...
+    /// Marks the beginning of the else branch.
+    Else,
+    /// Marks the end of the if statement.
+    EndIf,
 }
 
 impl Opcode {
     /// Take a byte slice and read the next opcode from it, including any associated data. `read_next` returns a tuple
     /// of the deserialised opcode, and an updated slice that has the Opcode and data removed.
     pub fn read_next(bytes: &[u8]) -> Option<(Opcode, &[u8])> {
```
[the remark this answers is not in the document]
delta1: will take a look

src/script/tari_script.rs
```diff
@@ -47,26 +52,76 @@ macro_rules! script {
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct TariScript {
     script: Vec<Opcode>,
+    state: ScriptState,
```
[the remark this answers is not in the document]
delta1: Will do

src/script/tari_script.rs
```diff
@@ -138,41 +196,193 @@ impl TariScript {
...
+            Add => TariScript::handle_op_add(stack),
+            Sub => TariScript::handle_op_sub(stack),
+            Equal => match TariScript::handle_equal(stack)? {
+                true => stack.push(Number(1)),
+                false => stack.push(Number(0)),
             },
-            Opcode::EqualVerify => match TariScript::handle_equal(stack)? {
-                true => stack.push(Number(0)),
+            EqualVerify => match TariScript::handle_equal(stack)? {
+                true => stack.push(Number(1)),
```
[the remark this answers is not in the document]
delta1: Good catch, thanks

src/script/tari_script.rs
```diff
@@ -47,26 +52,76 @@ macro_rules! script {
...
+        inputs: &ExecutionStack,
+        context: &ScriptContext,
+    ) -> Result<StackItem, ScriptError>
+    {
         // Copy all inputs onto the stack
         let mut stack = inputs.clone();
-        for opcode in self.script.iter() {
-            self.execute_opcode(opcode, &mut stack, context)?
+
+        let script = self.script.clone();
```
mikethetike, line 81: Perhaps `state` should rather be created as a local var here and then lives only as long as the execution

src/script/tari_script.rs
delta1: Oh I missed this one, did update in the other place Stan mentioned. My mistake

mikethetike (commented):
Looks good on the whole, but 2 comments and a nit

delta1:
Thanks for the feedback @CjS77 - on it

src/script/op_codes.rs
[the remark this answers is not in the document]
delta1: It's used in `TariScript::from_bytes`

src/script/tari_script.rs
delta1: This wasn't specified - do all `n + 1` items have to be of the same `StackItem` type?

src/script/op_codes.rs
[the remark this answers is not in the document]
delta1: making `read_next` private and exposing a public `parse` fn instead

src/script/tari_script.rs
delta1: Done

src/script/tari_script.rs
```diff
@@ -54,19 +58,65 @@ impl TariScript {
...
+    {
         // Copy all inputs onto the stack
         let mut stack = inputs.clone();
-        for opcode in self.script.iter() {
-            self.execute_opcode(opcode, &mut stack, context)?
+
+        // Local execution state
+        let mut state = ExecutionState::default();
+
+        let script = self.script.clone();
```
sdbondi, line 79: Believe there's no need to clone this now that execute_opcode is immutable.

CjS77 (approved):
LGTM

PR #37, opened 2021-03-09
Add time locked contract example for TariScript

philipr-za:
This PR adds an example for the time locked contract example from RFC-0202. It also update the `pay_to_public_key_hash` example to be inline with what is described in the RFC.


delta1:
Nice, just needs a fmt

philipr-za:
fmt and rebase

CjS77 (approved):
Nice.

PR #38, opened 2021-03-19
Add serde to the ExecutionStack

philipr-za:
Make storing the ExecutionStack in a database easier, this was manually implemented for TariScript but I think the derives are fine for the stack.

### DIFF
--- a/src/script/stack.rs
+++ b/src/script/stack.rs
@@ -31,7 +31,7 @@ pub const MAX_STACK_SIZE: usize = 256;
 #[macro_export]
 macro_rules! inputs {
     ($($input:expr),+) => {{
-        use crate::script::{ExecutionStack, StackItem};
+        use $crate::script::{ExecutionStack, StackItem};
 
         let items = vec![$(StackItem::from($input)),+];
         ExecutionStack::new(items)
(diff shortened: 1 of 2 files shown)


CjS77:
Where are you thinking of storing the stack in a DB?

philipr-za:
Storing an output's input data in the DB with the UnblindedOutput. I kept it as an execution stack so its clear what kind of "data" it is. I could just have it be `input_data: vec<u8>` but thought it was clearer as an ExecutionStack.

CjS77:
I suggest we roll our own `Vec<u8>` <-> `ExecutionStack` similar to how the script serialisation is done. It lets us make maximal use of the blockchain space with a compact binary representation.

This doesn't mean we can't also use serde in things like block explorers, but serde support could then be an opt-in feature.

philipr-za:
> I suggest we roll our own `Vec<u8>` <-> `ExecutionStack` similar to how the script serialisation is done. It lets us make maximal use of the blockchain space with a compact binary representation.
> 
> This doesn't mean we can't also use serde in things like block explorers, but serde support could then be an opt-in feature.

So ExecutionStack does have an `as_bytes` and `from_bytes` implementation already which are used for serialization for over-the-wire messages and when it is stored in the chain db. It's just much neater to have the Serde traits when working with Diesel and Sqlite.

I could definitely manually implement the serialize and deserialize traits using those functions but I don't see the benefit? The StackItems are just wrappers for the items inside them and the Serialize/Deserialize implementations on the Dalek types is doing the exact thing the manual `as_bytes` and `from_bytes` does, which is just calling the underlying as_bytes/from_bytes implementations?

CjS77 (approved):
LGTM

PR #43, opened 2021-06-10
Add commitment signature scheme using Ristretto

SWvheerden:
Added the commitment signature scheme for use in Tari crypto
Added unit tests covering the new scheme.
Co-Authored-By: Hansie Odendaal <pluto@tari.com>


src/lib.rs
```diff
@@ -11,7 +11,7 @@ pub mod hash;
 pub mod keys;
 pub mod musig;
 pub mod range_proof;
-pub mod signatures;
+pub mod signature;
```
mikethetike, line 14: These two mods with the same name is going to be confusing

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    /// This function returns the complete signature tuple (R, u, v)
+    #[inline]
+    pub fn get_complete_signature_tuple(&self) -> (&HomomorphicCommitment<P>, &K, &K) {
+        (&self.public_commitment_nonce, &self.signature_u, &self.signature_v)
+    }
+
+    /// This function returns the first publicly known private key of the signature tuple (u)
+    #[inline]
+    pub fn get_signature_u(&self) -> &K {
```
mikethetike, line 147:
```suggestion
    pub fn signature_u(&self) -> &K {
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+
+    /// This function returns the second publicly known private key of the signature tuple (v)
+    #[inline]
+    pub fn get_signature_v(&self) -> &K {
+        &self.signature_v
+    }
+
+    /// This function returns the public commitment nonce of the signature tuple (R)
+    #[inline]
+    pub fn get_public_commitment_nonce(&self) -> &HomomorphicCommitment<P> {
```
mikethetike, line 159:
```suggestion
    pub fn public_commitment_nonce(&self) -> &HomomorphicCommitment<P> {
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+#[derive(Clone, Debug, Error, PartialEq, Eq, Deserialize, Serialize)]
+pub enum CommitmentSignatureError {
+    #[error("An invalid challenge was provided")]
+    InvalidChallenge,
+}
+
+#[allow(non_snake_case)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CommitmentSignature<P, K, C> {
+    public_commitment_nonce: HomomorphicCommitment<P>,
```
mikethetike, line 50:
```suggestion
    nonce: HomomorphicCommitment<P>,
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+    signature_u: K,
```
mikethetike, line 51: I don't know if "signature" adds any value to the name here
```suggestion
    u: K,
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+    {
+        let lhs = self.calc_signature_verifier();
+        let rhs = &self.public_commitment_nonce + &(public_commitment * challenge);
+        // Implementors should make this a constant time comparison
+        lhs == rhs
+    }
+
+    /// This function returns the complete signature tuple (R, u, v)
+    #[inline]
+    pub fn get_complete_signature_tuple(&self) -> (&HomomorphicCommitment<P>, &K, &K) {
```
mikethetike, line 141: What is the value of this method?

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+    K: SecretKey,
+    C: HomomorphicCommitmentFactory<P = P> + Default,
+{
+    fn cmp(&self, other: &Self) -> Ordering {
+        match self
+            .get_public_commitment_nonce()
+            .cmp(&other.get_public_commitment_nonce())
+        {
+            Ordering::Equal => {
+                let this_u = self.get_signature_u().to_hex();
```
mikethetike, line 227: Is the hex conversion necessary here?

src/lib.rs
```diff
@@ -27,4 +27,5 @@ pub mod script;
 
 // Re-export tari_utils
 pub use hash::blake2 as common;
+pub use signature::schnorr as signatures;
```
mikethetike, line 30: disagree with this, rather just make a breaking change or expose it normally.

How about:
rename `pub mod signature` to `pub mod signatures`
and add
`pub use schorr::*` in `signatures.rs`

src/ristretto/ristretto_com_sig.rs
```diff
@@ -0,0 +1,224 @@
...
+        let nonce_commitment_alice = factory.commit(&k_1_alice, &k_2_alice);
+        // Alice generate some keys and nonces
+        let a_value_bob = RistrettoSecretKey::random(&mut rng);
+        let x_value_bob = RistrettoSecretKey::random(&mut rng);
+        let commitment_bob = factory.commit(&x_value_bob, &a_value_bob);
+        let k_1_bob = RistrettoSecretKey::random(&mut rng);
+        let k_2_bob = RistrettoSecretKey::random(&mut rng);
+        let nonce_commitment_bob = factory.commit(&k_1_bob, &k_2_bob);
+        // Each of them creates the Challenge committing to both commitments of both parties
+        let challange = Blake256::new()
```
mikethetike, line 194:
```suggestion
        let challenge = Blake256::new()
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+    signature_u: K,
+    signature_v: K,
+    _commitment_factory: PhantomData<C>,
+}
+
+impl<P, K, C> CommitmentSignature<P, K, C>
+where
+    P: PublicKey<K = K>,
+    K: SecretKey,
+    C: HomomorphicCommitmentFactory<P = P> + Default,
```
mikethetike, line 60: This can probably only be passed to the methods that need it. I suspect that the type of factory does not need to be specified as part of the type.

mikethetike (commented):
I need to test it locally. Looks good in general

src/lib.rs
SWvheerden: Trying to keep the old one not to break every program that uses tari_crypto. 
Changed the name to: `signature_schemes`

src/signature/commitment_signature.rs
SWvheerden: trying not to use single char names, but did update to use `u` and `v` rather

src/signature/commitment_signature.rs
SWvheerden: helper method so you dont have to call:
```
sig.u()
sig.v()
sig.nonce()
```

you can just call
`(_,_,_) = sig.get_complete_tuple`

src/lib.rs
SWvheerden: I saw the comment below, changing to just signatures, removing the old use.

src/signature/commitment_signature.rs
SWvheerden: its either a hex conversion or we need to impl ord with secret key

src/signature/commitment_signature.rs
CjS77: Note that by convention, Rust getters don't start with `get_`

src/signature/commitment_signature.rs
CjS77: 100% This is too opinionated. 

You've gone to the trouble of making the struct very general, but suddenly can't use this without default commitment factories.

Just pass in `&C` in the methods that need a factory, as Mike suggetsed.

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+        CommitmentSignature {
+            public_commitment_nonce,
+            signature_u,
+            signature_v,
+            _commitment_factory: PhantomData,
+        }
+    }
+
+    /// This is the left-hand side of the signature verification equation
+    pub fn calc_signature_verifier(&self) -> HomomorphicCommitment<P> {
```
CjS77:
```suggestion
    pub fn calc_signature_verifier(&self, factory: &C) -> HomomorphicCommitment<P> {
```

CjS77: The other approach is to calculate this in the constructor and store the value, providing this as a getter. Depends on how often the verifier is needed on its own

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+        challenge: &[u8],
+    ) -> Result<Self, CommitmentSignatureError>
+    where
+        K: Add<Output = K> + Mul<P, Output = P> + Mul<Output = K>,
+    {
+        let e = match K::from_bytes(challenge) {
+            Ok(e) => e,
+            Err(_) => return Err(CommitmentSignatureError::InvalidChallenge),
+        };
+        let ea = e.clone() * secret_a;
```
CjS77:
```suggestion
        let ea = &e* &secret_a;
```
will work I think, saving a clone.

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+        let ex = e * secret_x;
+
+        let v = nonce_a.clone() + ea;
+        let u = nonce_x.clone() + ex;
```
CjS77: ditto

src/signature/commitment_signature.rs
SWvheerden: That is a possibility, but it's  going to make the code messier. 

We use the signatures as: 
```
pub type RistrettoComSig = CommitmentSignature<RistrettoPublicKey, RistrettoSecretKey, PedersenCommitmentFactory>;
RistrettoComSig::sign(a_value_alice, x_value_alice, k_2_alice, k_1_alice, &challenge);
sig.verify_challenge(&commitment, &challenge));
```

If we remove the phantomdata

it's going to look as follows:
```
pub type RistrettoComSig = CommitmentSignature<RistrettoPublicKey, RistrettoSecretKey>;
RistrettoComSig::sign::<PedersenCommitmentFactory>(a_value, x_value, k_2, k_1, &challenge);
sig.verify_challenge::<PedersenCommitmentFactory>(&commitment, &challenge)
```

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+
+        let factory = C::default();
```
CjS77: Pass this in as an argument

src/signature/commitment_signature.rs
```diff
@@ -0,0 +1,292 @@
...
+        lhs == rhs
+    }
+
+    /// This function returns the complete signature tuple (R, u, v)
+    #[inline]
+    pub fn get_complete_signature_tuple(&self) -> (&HomomorphicCommitment<P>, &K, &K) {
+        (&self.public_commitment_nonce, &self.signature_u, &self.signature_v)
+    }
+
+    /// This function returns the first publicly known private key of the signature tuple (u)
```
CjS77: The way I read this is that `u` is a partial signature rather than the oddly worded public private key. Is this how the paper refers to it?

CjS77 (commented):
Generally looks good.

The pinning of the default commitment factory needs to be addressed though.

A utility struct for keeping the number of arguments to the functions might help, or
keep a reference to the factory in the struct (with a suitable lifetime) -- this is preferred since it ensures that you use the same factory throughout.

Or use the builder pattern (less preferred -- this is still pretty simple).

src/signature/commitment_signature.rs
SWvheerden: Yeah, the paper refers to this as u and v. Hence why we decided to go with u and v as the names to them.

src/signature/commitment_signature.rs
SWvheerden: I prefer locking it down. Because the `factory` will be related to the `HomomorphicCommitment` which is in turn related to the `public/private keys` Its not something that can change midway. You need to ensure you use the correct one to sign and verify. 

As to the reference, and or using a generic on only the function. Because of the way the code is written, you will need to pass that in every time you sign and or verify. As `calc_signature_verifier` is called by `verify` which is used by `verify_challenge`.  The only functions that don't use the factory are the getters.
So if you need to pass one in every time, why not just lock it down at creation?

src/ristretto/musig.rs
```diff
@@ -680,8 +680,8 @@ mod test {
             r_agg = r_agg + r;
         }
         assert_eq!(musig.has_failed(), false);
-        if msg.is_some() {
-            musig = musig.set_message(msg.unwrap());
+        if let Some(v) = msg {
```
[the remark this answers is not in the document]
SWvheerden: Yeah this was just to get clippy to pass on the repo

src/signature/commitment_signature.rs
mikethetike: surely the hex has the same order as sorting the bytes?

src/wasm/key_utils.rs
```diff
@@ -169,7 +314,7 @@ pub(crate) fn sign_with_key(k: &RistrettoSecretKey, e: &[u8], r: Option<&Ristret
 /// Checks the validity of a Schnorr signature
 #[allow(non_snake_case)]
 #[wasm_bindgen]
-pub fn check_signature(pub_nonce: &str, signature: &str, pub_key: &str, msg: &str) -> JsValue {
+pub fn check_comsig_signature(pub_nonce: &str, signature: &str, pub_key: &str, msg: &str) -> JsValue {
```
mikethetike, line 317: Is this correct?

src/wasm/key_utils.rs
```diff
@@ -91,7 +110,127 @@ pub fn pubkey_from_secret(k: &str) -> Option<String> {
...
+
+    let sig = RistrettoSchnorr::new(R, s);
+    let msg = Blake256::digest(msg.as_bytes());
+    result.result = sig.verify_challenge(&P, msg.as_slice());
+    JsValue::from_serde(&result).unwrap()
+}
+
+/// Generate a Commitment signature of the message using the given private key
+#[wasm_bindgen]
+pub fn sign_comsig(private_key: &str, msg: &str) -> JsValue {
```
mikethetike, line 233: Is this correct?

mikethetike (commented):
See it's still WIP, but added these comments in case they get missed

mikethetike (approved):

PR #46, opened 2021-06-30
Add `to_vec` function to comm sig

sdbondi:
This implements the canonical way to encode CommitmentSignature.
i.e `bytes(R) || bytes(u) || bytes(v)`

Unfortunately, implementing `ByteArray` is not possible (maybe with unsafe) as things stand 
because the trait requires `as_bytes` to be implemented which implies 
something owns the contiguous bytes that the slice is referencing, which is not the case.

## Usage

The script_signature needs to be committed to in the input hash. ~~One way to do that is to calculate `uG + vH` where `u` and `v` are the public signatures (i.e use  `calc_signature_verifier(&CommitmentFactory::default()).as_bytes()`. However, this requires all clients to be able to create commitments on G and H (extra wasm bindings etc) vs simply concatenating bytes.~~ EDIT: was obviously tired ;) can  just use `bytes(u)`, `bytes(v)` and `bytes(nonce)`. Perhaps this is still useful, otherwise feel free to close.

### DIFF
--- a/src/ristretto/ristretto_com_sig.rs
+++ b/src/ristretto/ristretto_com_sig.rs
@@ -102,10 +102,11 @@ mod test {
     use crate::{
         commitment::HomomorphicCommitmentFactory,
         common::Blake256,
-        keys::SecretKey,
+        keys::{PublicKey, SecretKey},
         ristretto::{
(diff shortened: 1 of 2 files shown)


delta1 (approved):
LGTM

CjS77 (approved):
lgtm

PR #47, opened 2021-07-02
Stable rust

CjS77:
Updated dependencies such that Rust stable 1.53 is now supported.

The optimised `avx_2` option will NOT run on stable because there's still an unstable feature on `subtle-ng`. BUT this feature is actually for doc generation and has been removed from Rust. As soon as `subtle-ng`
merges dalek-cryptography/subtle#85,`avx2` will probably be supported on stable as well.

* Update benchmark code to use new dependency APIs
* Run rust-fmt


delta1 (approved):
LGTM, untested

PR #48, opened 2021-07-02
Source coverage

CjS77:
## Use source-bsed code coverage, rather than gcov

This PR replaces the grcov action with the more accurate source coverage approach as detailed in
[the Rust unstable book](https://doc.rust-lang.org/nightly/unstable-book/compiler-flags/instrument-coverage.html)
    
This PR also provides the `test_coverage.sh` script that generates a test coverage report locally (see the comments for setting up  pre-requisites).

### Before

Screenshot 2021-07-02 at 15 21 15

### After

Screenshot 2021-07-02 at 15 44 01


test_coverage.sh
```diff
@@ -0,0 +1,49 @@
...
+    --show-branch-summary \
+    --show-instantiation-summary \
+    --show-region-summary \
+    --ignore-filename-regex='/.cargo/registry' \
+    --instr-profile=cov_raw/tari_crypto.profdata \
+    $files \
+    > cov_raw/tari_crypto.lcov
+
+if [ -z ${SKIP_HTML+x} ]; then
+  genhtml -o coverage_report cov_raw/tari_crypto.lcov
```
sdbondi, line 45: I got this error in the report generation on mac:

`genhtml: ERROR: cannot read /rustc/657bc01888e6297257655585f9c475a0801db6d2/library/std/src/thread/local.rs`

I added `--ignore-filename-regex="^/rustc" \` to the `cargo cov` command  which worked - not sure if that is fixing a symptom though.

sdbondi (dismissed):
Looks great - tested and approved because the error I got has more to do with local report generation than coveralls.

test_coverage.sh
CjS77: Thanks; I think that exclusion makes sense.

sdbondi (approved):

Issue #34, opened 2021-03-02
Blake2 dep is outdated, updating it has some issues to fix 


delta1:
Blake2 0.8.1 relies on crypto-mac 0.7.0 > generic-array 0.12.3 which has the following rust-sec advisory: 

https://rustsec.org/advisories/RUSTSEC-2020-0146.html 

Bumping to Blake2 0.9.1 leads to some compilation issues to be fixed. Will take a look when I can. 

Context: tari-project/tari#2714

CjS77:
Fixed in #49

PR #49, opened 2021-07-03
Migrate to digest 0.9

### LINKED ISSUE #35: RUSTSEC-2020-0146: arr! macro erases lifetimes
> arr! macro erases lifetimes

| Details             |                                                |
| ------------------- | ---------------------------------------------- |
| Package             | `generic-array`                      |
| Version             | `0.12.4`                   |
| URL                 | [https://github.com/fizyk20/generic-array/issues/98](https://github.com/fizyk20/generic-array/issues/98) |
| Date                | 2020-04-09                         |
| Patched versions    | `>=0.14.0`                  |
| Unaffected versions | `<0.8.0`               |

Affected versions of this crate allowed unsoundly extending
lifetimes using `arr!` macro. This may result in a variety of
memory corruption scenarios, most likely use-after-free.

See [advisory page](https://rustsec.org/advisories/RUSTSEC-2020-0146.html) for additional details.

CjS77:
This PR moves all dependencies to use the digest 0.9 traits and APIs.
This is a breaking change, so the minor version is incremented.

Clients of this generally only need to update the `result` method to
`finalize`; and obviously make use of the v0.9 `digest::Digest` trait
where necessary.

As a result, the deprecated k12, sha3 and Blake3 objects can be removed.
Methods and functions that need a hasher are all generic over `Digest`.

We retain the convenience wrapper over `VarBlake2B` to produce 256 bit
hashes and implement the necessary sub-traits to support
`digest::Digest`.

This update also fixes #35 and #34


src/ristretto/ristretto_keys.rs
```diff
@@ -256,9 +254,7 @@ impl DiffieHellmanSharedSecret for RistrettoPublicKey {
 // Requires custom Hashable implementation for RistrettoPublicKey as CompressedRistretto doesnt implement this trait
 impl Hashable for RistrettoPublicKey {
     fn hash(&self) -> Vec<u8> {
-        let mut hasher = HashDigest::new();
-        hasher.input(&self.to_vec());
-        hasher.result().to_vec()
+        Blake2b::digest(&self.to_vec()).to_vec()
```
sdbondi, line 257:
```suggestion
        Blake2b::digest(self.as_bytes()).to_vec()
```

sdbondi (commented):
Nice! One minor change

sdbondi (approved):

PR #51, opened 2021-07-08
Minor clippy fixes

sdbondi:
- [minor perf] Use count instead of collect in `pop_n_plus_one_contains`
- [test] Use `PartialEq` impl instead of `matches!` to check non-static
  `Opcode` 

~~Not very intuitively, `assert!(matches!(opcode, op))` is producing a `match opcode` and `op` is placed in the catch-all position, so not testing the match at all~~ see comments

### DIFF
--- a/src/script/stack.rs
+++ b/src/script/stack.rs
@@ -212,14 +212,14 @@ impl ExecutionStack {
 
         // check that all popped items are of the same variant
         // first count each variant
-        let counts = items.iter().fold([0; 5], |values, item| counter(values, item));
+        let counts = items.iter().fold([0; 5], counter);
         // also check the n + 1 item
         let counts = counter(counts, &item);
 
         // then filter those with more than 0
-        let distict_variants: Vec<&u8> = counts.iter().filter(|&c| c > &0).collect();
+        let num_distinct_variants = counts.iter().filter(|&c| *c > 0).count();
 
-        if distict_variants.len() > 1 {
+        if num_distinct_variants > 1 {
             return Err(ScriptError::InvalidInput);
         }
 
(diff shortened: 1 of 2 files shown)


delta1:
> `op` is placed in the catch-all position 

Wouldn't `op` be the only match besides the catch-all which returns false? 

https://doc.rust-lang.org/src/core/macros/mod.rs.html#227-234

src/script/stack.rs
```diff
@@ -212,14 +212,14 @@ impl ExecutionStack {
...
         // check that all popped items are of the same variant
         // first count each variant
-        let counts = items.iter().fold([0; 5], |values, item| counter(values, item));
+        let counts = items.iter().fold([0; 5], counter);
         // also check the n + 1 item
         let counts = counter(counts, &item);
 
         // then filter those with more than 0
-        let distict_variants: Vec<&u8> = counts.iter().filter(|&c| c > &0).collect();
+        let num_distinct_variants = counts.iter().filter(|&c| *c > 0).count();
```
delta1, line 220: nice!

delta1 (approved):

sdbondi:
> Wouldn't op be the only match besides the catch-all which returns false?

Ah you are correct, so it is testing it but to resolve the clippy we need to use `matches!(opcode, _op)`

```rust 
match opcode {
          op => true,
          _ => false
}
```

delta1:
Cool, what warning was clippy throwing?

sdbondi:
> Cool, what warning was clippy throwing?
Unused var `op`

delta1:
oh lol that's annoying 😓

CjS77 (approved):

PR #52, opened 2021-07-08
Remove leaky mock and remove libc dependency

sdbondi:
- Ristretto random_keypair overrides in FFI key tests leak and cause
  other tests to fail when `ffi` feature is enabled. The cause of the
  failure can be very difficult to track down. Changed the tests to
  assert that the key was changed to "something else".
- Remove `libc` because stdlib has the `c_char` type 
- Added `--all-features` flag to ci


CjS77 (dismissed):
LGTM

mikethetike (approved):

PR #54, opened 2021-07-18
Minor optimisation to SchnorrSignature ord impl

sdbondi:
Use little-endian byte order (32 bytes comparison, 0 allocations) instead of
hex order (64 byte comparison + 2 allocations). These comparisons are
equivalent (NB because this impl is consensus code).


delta1 (approved):
Nice one LGTM, untested

CjS77 (approved):

PR #56, opened 2021-09-03
fix: remove extra compress call during pubkey::deserialize

mikethetike:
During deserialization, the compress and decompress method is called, when only one of them needs to be.

This should halve the time during `from_bytes()`

image

### DIFF
--- a/src/ristretto/ristretto_keys.rs
+++ b/src/ristretto/ristretto_keys.rs
@@ -213,11 +213,17 @@ pub struct RistrettoPublicKey {
 
 impl RistrettoPublicKey {
     // Private constructor
-    pub(crate) fn new_from_pk(pk: RistrettoPoint) -> RistrettoPublicKey {
+    pub(super) fn new_from_pk(pk: RistrettoPoint) -> RistrettoPublicKey {
         RistrettoPublicKey {
             point: pk,
             compressed: pk.compress(),
         }
+    }
+
+    pub(super) fn new_from_compressed(compressed: CompressedRistretto) -> Option<RistrettoPublicKey> {
+        compressed
+            .decompress()
+            .map(|point| RistrettoPublicKey { compressed, point })
     }
 }
 
@@ -332,12 +338,12 @@ impl ByteArray for RistrettoPublicKey {
         if bytes.len() != 32 {
             return Err(ByteArrayError::IncorrectLength);
         }
-        let pk = CompressedRistretto::from_slice(bytes);
-        match pk.decompress() {
+        let compressed = CompressedRistretto::from_slice(bytes);
+        match RistrettoPublicKey::new_from_compressed(compressed) {
+            Some(p) => Ok(p),
             None => Err(ByteArrayError::ConversionError(
                 "Invalid compressed Ristretto point".to_string(),
             )),
-            Some(p) => Ok(RistrettoPublicKey::new_from_pk(p)),
         }
     }
 


delta1 (approved):
LGTM untested

src/ristretto/ristretto_keys.rs
```diff
@@ -219,6 +219,13 @@ impl RistrettoPublicKey {
             compressed: pk.compress(),
         }
     }
+
+    pub(crate) fn new_from_compressed(compressed: CompressedRistretto) -> Option<RistrettoPublicKey> {
+        let c = compressed.clone();
+        compressed
+            .decompress()
+            .map(|point| RistrettoPublicKey { compressed: c, point })
+    }
```
sdbondi, line 228:
```suggestion
        compressed
            .decompress()
            .map(|point| RistrettoPublicKey { compressed, point })
    }
```

src/ristretto/ristretto_keys.rs
```diff
@@ -219,6 +219,13 @@ impl RistrettoPublicKey {
             compressed: pk.compress(),
         }
     }
+
+    pub(crate) fn new_from_compressed(compressed: CompressedRistretto) -> Option<RistrettoPublicKey> {
```
CjS77, line 223: Feel like this should return a `Result` since our contract is that we always store a valid compressed/decompressed pair.

CjS77 (commented):
Nice optimisation. One nit.

sdbondi (commented):
Ran some dalek criterion benches and looks like decompress and compress times are similar. The implementation of decompress uses constant time comparison, so could be a little slower than compress.   

```
RistrettoPoint decompression                                                                             
                        time:   [5.7858 us 5.8601 us 5.9463 us]
                        change: [-3.5477% -2.4195% -1.2709%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) high mild
  1 (1.00%) high severe

RistrettoPoint compression                                                                             
                        time:   [4.0617 us 4.0944 us 4.1341 us]
                        change: [-52.136% -48.548% -45.005%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  9 (9.00%) high mild
  1 (1.00%) high severe
```

src/ristretto/ristretto_keys.rs
mikethetike: Cool, agree

src/ristretto/ristretto_keys.rs
mikethetike: Would this even compile? Will try

src/ristretto/ristretto_keys.rs
CjS77: `map(..).ok_or(E)` will do it

mikethetike: There isn't an Error struct yet in the crate for this kind of error, and the method is only called privately from Deserialize which already exposes a ByteArrayError, so I restricted it further to `super` to reflect that it is private only and kept it as the `Option` that dalek exposes.

src/ristretto/ristretto_keys.rs
mikethetike: Applied. I thought it was a move, but it's a borrow

sdbondi (approved):
Looks good!

PR #59, opened 2021-09-14
feat: add CheckMultiSig and CheckMultiSigVerify

delta1:
- found and fixed a bug in the if-then-else implementation: when the "then" branch of an if-then-else is run, then nested
if-then-else statements in the "else" branch cause an "invalid opcode" error for a valid script
- add implementation and tests for `CheckMultiSig` and `CheckMultiSigVerify` opcodes


delta1:
`rust-toolchain` is specifying `nightly-2021-06-01` but circleci is using `nightly-2020-09-11`

CjS77 (commented):
Surely this would all be much simpler if branches were handled with recursion, something like

```
fn execute_script(stack, script) {
  POP opcode -> op
  match op {
     ..
     IFTHEN => 
       extract_clauses(stack, script) -> (pred,tc,ec)
       if pred
       then execute_script(stack, tc)
       else execute_script(stack, ec)
    ...
```

Then you never have to worry about branching logic again.

Caveat - We'd need a MAXIMUM_DEPTH parameter to guard against deep recursion bugs

mikethetike (dismissed):
Looks good. Only reviewed the multsig stuff, not the if/else.

Question on the message/challenge. Is it always going to be known when the script is created? Can it be read from the stack?

CjS77 (commented):
I strongly suggest we force m,n, and the set of pubkeys to be in the script as parameters to the opcode and not pop them off the stack.

Even though we give up a little flexibility in doing this, the benefits are
* Simpler to construct scripts.
* Multisig counts as one opcode and not 4+
* Remove an entire class of bugs stemming from poorly written scripts that inadvertently allow the spender to control the multisig parameters.

I'd reckon 98% of the time, we know m,n and the pubkeys apriori and can put them in the script.
I'm allowing 2% margin of error, because I think it might be 100%

If there are use cases for putting these parameters on the stack, we can add an `UnsfaeMultisig` opcode to accommodate it (with the intentionally scary name pointing people to the safer version)

src/script/op_codes.rs
```diff
@@ -225,6 +227,16 @@ pub enum Opcode {
     /// Identical to CheckSig, except that nothing is pushed to the stack if the signature is valid, and the operation
     /// fails with VERIFY_FAILED if the signature is invalid.
     CheckSigVerify(Box<Message>),
+    /// Pop the number `n` indicating the number of allowed spenders, and then pop `n` spender public keys off the
+    /// stack. Then pop the number `m` indicating the number of required spending signatures, and then pop `m`
+    /// signatures from the stack provided as inputs to the script. If `m` signatures of the possible `n` public keys
+    /// sign the challenge message, push 1 to the stack, otherwise push 0. Fails with INVALID_SCRIPT_DATA if the Msg is
+    /// not a valid 32-byte value. Fails with EMPTY_STACK if the stack has too few items. Fails with INVALID_INPUT
+    /// if the popped elements are not of the correct types, of if `m` or `n` are 0 or if `m` is greater than `n`.
+    CheckMultiSig(Box<Message>),
```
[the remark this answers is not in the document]
delta1: Sure sounds good

delta1:
> Looks good. Only reviewed the multsig stuff, not the if/else.
> 
> Question on the message/challenge. Is it always going to be known when the script is created? Can it be read from the stack?

I presume it could be read from the stack, but was copying the way `CheckSig` worked with the message supplied in the script.

delta1:
> Surely this would all be much simpler if branches were handled with recursion, something like
> 
> ```
> fn execute_script(stack, script) {
>   POP opcode -> op
>   match op {
>      ..
>      IFTHEN => 
>        extract_clauses(stack, script) -> (pred,tc,ec)
>        if pred
>        then execute_script(stack, tc)
>        else execute_script(stack, ec)
>     ...
> ```
> 
> Then you never have to worry about branching logic again.
> 
> Caveat - We'd need a MAXIMUM_DEPTH parameter to guard against deep recursion bugs

That would be more elegant, but I think the devil is in the `extract_clauses` detail there. 
I disagree with changing this now since at best it's a refactor just to get back to this point, at worst it could introduce new bugs. 

Anyway I'm happy to try it out if you feel strongly about it LMK.

src/script/op_codes.rs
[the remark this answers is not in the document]
hansieodendaal: > What we want is this:
> 

Agreed, to should be similar to the [Bitcoin m-of-n multisig](https://tlu.tarilabs.com/protocols/mimblewimble-mp-bp-utxo/MainReport.html#bitcoin--mtext-of-n--multisig-in-a-nutshell).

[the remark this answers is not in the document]
delta1: @hansieodendaal thats exactly how this current version works.

CjS77:
> > Surely this would all be much simpler if branches were handled with recursion, something like
> > ```
> > fn execute_script(stack, script) {
> >   POP opcode -> op
> >   match op {
> >      ..
> >      IFTHEN => 
> >        extract_clauses(stack, script) -> (pred,tc,ec)
> >        if pred
> >        then execute_script(stack, tc)
> >        else execute_script(stack, ec)
> >     ...
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then you never have to worry about branching logic again.
> > Caveat - We'd need a MAXIMUM_DEPTH parameter to guard against deep recursion bugs
> 
> That would be more elegant, but I think the devil is in the `extract_clauses` detail there.
> I disagree with changing this now since at best it's a refactor just to get back to this point, at worst it could introduce new bugs.
> 
> Anyway I'm happy to try it out if you feel strongly about it LMK.

 iirc it was originally written recursively and that the current approach has had more than one PR to fix bugs in the if-then-else logic.

CjS77:
> > Looks good. Only reviewed the multsig stuff, not the if/else.
> > Question on the message/challenge. Is it always going to be known when the script is created? Can it be read from the stack?
> 
> I presume it could be read from the stack, but was copying the way `CheckSig` worked with the message supplied in the script.

Yeah, there should be a tendency to try and lock things in at script/UTXO creation time to reduce attack surface.

The exceptions are 
* where the parameter must _obviously_ come at spending time.
* where flexibility is reduced to the point where 80% of use cases are removed by locking parameters in at script time.

delta1:
> > > Surely this would all be much simpler if branches were handled with recursion, something like
> > > ```
> > > fn execute_script(stack, script) {
> > >   POP opcode -> op
> > >   match op {
> > >      ..
> > >      IFTHEN => 
> > >        extract_clauses(stack, script) -> (pred,tc,ec)
> > >        if pred
> > >        then execute_script(stack, tc)
> > >        else execute_script(stack, ec)
> > >     ...
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > Then you never have to worry about branching logic again.
> > > Caveat - We'd need a MAXIMUM_DEPTH parameter to guard against deep recursion bugs
> > 
> > 
> > That would be more elegant, but I think the devil is in the `extract_clauses` detail there.
> > I disagree with changing this now since at best it's a refactor just to get back to this point, at worst it could introduce new bugs.
> > Anyway I'm happy to try it out if you feel strongly about it LMK.
> 
> iirc it was originally written recursively and that the current approach has had more than one PR to fix bugs in the if-then-else logic.

Nope. I wrote the first implementation and this is the first (only) bug fix. If there are other bugs I’ll gladly rewrite it.

src/script/tari_script.rs
```diff
@@ -428,6 +461,55 @@ impl TariScript {
...
+            .pop_num_items(m)?
+            .into_iter()
+            .map(|item| match item {
+                Signature(s) => Ok(s),
+                _ => Err(ScriptError::IncompatibleTypes),
+            })
+            .collect::<Result<Vec<RistrettoSchnorr>, ScriptError>>()?;
+
+        let mut valid_sigs = 0;
+        for signature in signatures.iter() {
```
mikethetike, line 502: Doesn't this mean that one public key could sign the message n times and it would be valid? We should remove the valid public keys from the list after it's been verified

delta1: Good catch

delta1:
- added github actions
- fixed bug (thanks @mikethetike) 
- changed CheckMultiSig/Verify to take data in args instead of from stack 
- added max N limit of 32 
- updated the `script!` macro to accept opcodes with more than 1 arg

src/script/tari_script.rs
```diff
@@ -428,6 +467,42 @@ impl TariScript {
...
+        let mut key_signed = vec![false; public_keys.len()];
+        for (i, public_key) in public_keys.iter().enumerate() {
+            for signature in signatures.iter() {
+                if !key_signed[i] && signature.verify_challenge(public_key, &message) {
+                    valid_signatures += 1;
+                    key_signed[i] = true;
+                    break;
+                }
+            }
+        }
```
CjS77, line 502: This catches the issue where one key submits 2 sigs, but I don't think it solves the case @mikethetike raised?

Best to add a test for submitting 3 of 5 multisig where pubkey 1 and 3 are repeated, and sig 3 is signed by that key.

CjS77 (commented):
There still looks to be a bug in the check sig logic

src/script/tari_script.rs
delta1: Should the supplied public keys be enforced to be unique?

src/script/op_codes.rs
```diff
@@ -351,6 +387,27 @@ impl Opcode {
...
+        let num = *n as usize;
+        let len = 3 + num * PUBLIC_KEY_LENGTH;
+        if bytes.len() < len {
+            return Err(ScriptError::InvalidData);
+        }
+        let keys = slice_to_vec_pubkeys(&bytes[3..len], num)?;
+        let end = len + 32;
+        if bytes.len() < end {
+            return Err(ScriptError::InvalidData);
+        }
```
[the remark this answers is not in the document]
delta1: `&bytes[len..end]` will panic if `end` is out of range for the slice?

src/script/tari_script.rs
CjS77: We can allow duped pubkeys, but we should enforce that each signature is used on at most one pubkey, and that each pubkey is used at most once.

A simpler and more direct test would be
MultiSig(2, 3, [P_a, P_b, P_a]) with [s_a, s_x] should fail (s_a cannot be used on both P_as to get 2 valid signatures)

src/script/tari_script.rs
```diff
@@ -462,53 +468,40 @@ impl TariScript {
...
             .collect::<Result<Vec<RistrettoSchnorr>, ScriptError>>()?;
 
-        let mut valid_sigs = 0;
-        for signature in signatures.iter() {
-            for public_key in public_keys.iter() {
-                if signature.verify_challenge(public_key, &message) {
-                    valid_sigs += 1;
+        let mut valid_signatures = 0;
+        let mut key_signed = vec![false; public_keys.len()];
+        for (i, public_key) in public_keys.iter().enumerate() {
```
mikethetike: I suggest switching these two loops, then you can fail fast as soon an invalid signature is encountered.

src/script/tari_script.rs
```diff
@@ -462,53 +468,40 @@ impl TariScript {
...
+            for signature in signatures.iter() {
+                if !key_signed[i] && signature.verify_challenge(public_key, &message) {
+                    valid_signatures += 1;
+                    key_signed[i] = true;
                     break;
                 }
             }
```
mikethetike, line 503: Add a fast fail here.
```suggestion
            }
            if !any_valid {
              return Err(Invalid)
            }
```

Looping over the signatures first means you are effectively removing the valid signature and the pubkey used from the set whenever one is found. If you leave the pub key as the outer loop, both need to be removed from the set manually

src/script/tari_script.rs
delta1: Just added that fix and a test for it

src/script/op_codes.rs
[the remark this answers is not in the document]
CjS77: You should check for len+32 in the first check

src/script/tari_script.rs
delta1: updated the test to add your test case: [link](https://github.com/tari-project/tari-crypto/pull/59/commits/31a1e5d4a4456bee9e1d9ee04f32b2ff51eabd5c#diff-2b1d02202ea28e771ea738d8c773f3dee0f3d216c5afdd6c613781709f97b33eR1071)

src/script/tari_script.rs
delta1: good idea

Cargo.toml
```diff
@@ -17,8 +17,8 @@ digest = "0.9.0"
 rand = { version = "0.8", default-features = false }
 getrandom = { version = "0.2.3", default-features = false, optional = true }
 clear_on_drop = "=0.2.4"
-curve25519-dalek = { package = "curve25519-dalek-ng", version = "4", default-features = false, features = ["u64_backend", "serde", "alloc"] }
-bulletproofs = {version = "4.0.0", package="tari_bulletproofs"}
+curve25519-dalek = { package = "curve25519-dalek-ng", version = "4.1", default-features = false, features = ["u64_backend", "serde", "alloc"] }
+bulletproofs = { version = "4.1", package="tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", branch = "main" } # todo: update when 4.1 pushed to crates.io
```
delta1, line 21: can we push tari_bulletproofs 4.1 to crates.io then this can be updated

CjS77 (dismissed):
LGTM

src/script/op_codes.rs
```diff
@@ -56,6 +59,21 @@ pub fn slice_to_boxed_message(slice: &[u8]) -> Box<Message> {
...
+        return Err(ScriptError::InvalidData);
+    }
+
+    let public_keys = slice
+        .chunks_exact(PUBLIC_KEY_LENGTH)
+        .take(num)
+        .map(RistrettoPublicKey::from_bytes)
+        .collect::<Result<Vec<RistrettoPublicKey>, ByteArrayError>>()?;
+
+    Ok(public_keys)
```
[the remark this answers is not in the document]
delta1: Clippy actually didn't mind this one 🤷 

Personally, I prefer `Ok(public_keys)` to 

```
    Ok(slice
        .chunks_exact(PUBLIC_KEY_LENGTH)
        .take(num)
        .map(RistrettoPublicKey::from_bytes)
        .collect::<Result<Vec<RistrettoPublicKey>, ByteArrayError>>()?)
```

delta1:
- added cargo fmt and clippy checks to PRs 
- fixed all the clippy warnings 🖇

mikethetike (approved):

src/script/op_codes.rs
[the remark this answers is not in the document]
CjS77: Doesn't 
```
slice
        .chunks_exact(PUBLIC_KEY_LENGTH)
        .take(num)
        .map(RistrettoPublicKey::from_bytes)
        .collect::<Result<Vec<RistrettoPublicKey>, ByteArrayError>>()
```
work?

PR #60, opened 2021-09-15
fix: reduce memory allocation on execution_stack and scripts

hansieodendaal:
This PR is to combat a memory leak for the console wallet in Ubuntu 18.04 and 20.04. 

Instrumenting the console wallet with https://docs.rs/dhat/0.2.2/dhat/ and testing with `valgrind --tool=dhat` highlighted `Script` and `ExecutionStack` as the culprits. Pre-allocation of `Script` and `ExecutionStack` vectors (`Vec::with_capacity(512)`) resulted in memory usage of 12GB vs. 4.4GB when using `Vec::new()` for a large wallet.

View with https://nnethercote.github.io/dh_view/dh_view.html

**With this PR**
[dhat-heap.json.30285_vec_new.log](https://github.com/tari-project/tari-crypto/files/7167562/dhat-heap.json.30285_vec_new.log)

**Pre this PR**
[dhat-heap.json.19190.log](https://github.com/tari-project/tari-crypto/files/7167554/dhat-heap.json.19190.log)

### DIFF
--- a/src/script/op_codes.rs
+++ b/src/script/op_codes.rs
@@ -270,7 +270,7 @@ pub enum Opcode {
 
 impl Opcode {
     pub fn parse(bytes: &[u8]) -> Result<Vec<Opcode>, ScriptError> {
-        let mut script = Vec::with_capacity(512);
+        let mut script = Vec::new();
         let mut bytes_copy = bytes;
 
         while !bytes_copy.is_empty() {
--- a/src/script/stack.rs
+++ b/src/script/stack.rs
@@ -243,14 +243,14 @@ impl ExecutionStack {
 
     /// Return a binary array representation of the input stack
     pub fn as_bytes(&self) -> Vec<u8> {
-        self.items.iter().fold(Vec::with_capacity(512), |mut bytes, item| {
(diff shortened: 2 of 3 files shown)


CjS77:
It's generally good practice to use `with_capacity` if you know a reasonable upper bound of allocation size. `Vec::new()` allocates a length of zero, meaning that you will get at least one additional heap allocation on the next push.

What is the median and  95th percentile of script length?

Maybe 512 is too high here, and we can reduce that value, but
a) `Vec::new()` will hurt performance if most scripts are non-zero length.
b) This doesn't feel like the root cause of a memory leak, where memory allocation is growing without bound.

It does point to some Map or Vec of UTXOs that is growing though, since that's one place where `Script` is stored.

CjS77 (approved):

hansieodendaal:
> It does point to some Map or Vec of UTXOs that is growing though, since that's one place where `Script` is stored.

Agreed that this does not fix the memory leak in the Ubuntu console wallet, but only optimize the general memory usage and performance. I will follow the code paths in the console wallet to find the root cause.

mikethetike (approved):

PR #65, opened 2021-11-16
fix: ensure ExecutionStack cannot exceed MAX_STACK_SIZE

sdbondi:
- Check that ExecutionStack does not exceed MAX_STACK_SIZE (255) when decoding from bytes
- Add `size()` to tari script ~~required to limit the maximum number of script elements in transaction and block validation~~

### DIFF
--- a/src/script/stack.rs
+++ b/src/script/stack.rs
@@ -26,7 +26,7 @@ use tari_utilities::{
     hex::{from_hex, to_hex, Hex, HexError},
     ByteArray,
 };
-pub const MAX_STACK_SIZE: usize = 256;
+pub const MAX_STACK_SIZE: usize = 255;
 
 #[macro_export]
 macro_rules! inputs {
@@ -250,18 +250,18 @@ impl ExecutionStack {
     }
 
     pub fn from_bytes(bytes: &[u8]) -> Result<Self, ScriptError> {
-        let mut items = Vec::new();
+        let mut stack = ExecutionStack { items: Vec::new() };
         let mut byte_str = bytes;
         while !byte_str.is_empty() {
             match StackItem::read_next(byte_str) {
                 Some((item, b)) => {
-                    items.push(item);
+                    stack.push(item)?;
                     byte_str = b;
                 },
                 None => return Err(ScriptError::InvalidInput),
             }
         }
-        Ok(ExecutionStack { items })
+        Ok(stack)
     }
 
     /// Pushes the item onto the top of the stack. This function will only error if the new stack size exceeds the
--- a/src/script/tari_script.rs
+++ b/src/script/tari_script.rs
@@ -99,6 +99,11 @@ impl TariScript {
         }
     }
 
+    /// Returns the number of script op codes
+    pub fn size(&self) -> usize {
+        self.script.len()
+    }
+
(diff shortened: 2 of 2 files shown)


src/script/stack.rs
```diff
@@ -250,24 +250,24 @@ impl ExecutionStack {
...
         }
-        Ok(ExecutionStack { items })
+        Ok(stack)
     }
 
     /// Pushes the item onto the top of the stack. This function will only error if the new stack size exceeds the
     /// maximum allowed stack size, given by [MAX_STACK_SIZE]
     pub fn push(&mut self, item: StackItem) -> Result<(), ScriptError> {
-        if self.size() >= MAX_STACK_SIZE {
+        if self.size() > MAX_STACK_SIZE {
```
CjS77, line 270:
```suggestion
        if self.size() >= MAX_STACK_SIZE {
```
because we're about to increase the size by one, which would exceed MAX SIZE if it were equal.

CjS77 (commented):
LGTM except for the off-by-one

src/script/stack.rs
sdbondi: This one was actually more about the docs/error message (I decreased MAX_STACK_SIZE by one)
In the comments it says if it _exceeds_ [MAX_STACK_SIZE] which was 256 - meaning that it can be 256, but in reality 255 is the maximum it can contain.

sdbondi:
@CjS77 What are your thoughts on limiting the number of ops per script in tari crypto? 
Looking at bitcoin, there is a 10000 byte limit on the script byte size, a limit of 520 items on the input stack, 1000 limit on items pushed by a script on the stack ([script.h](https://github.com/bitcoin/bitcoin/blob/master/src/script/script.h#L24)). Perhaps we should introduce similar limits?

CjS77:
We have an implicit limit on ops since we limit the script size, don't we?

src/script/stack.rs
CjS77: The original looks consistent to me. 

"This function will only error if the **new stack size** exceeds the maximum allowed stack size, given by [MAX_STACK_SIZE]"

MAX_STACK_SIZE should be whatever. And that's the maximum stack size, not MAX_STACK_SIZE+1. I figured you changed MAX_STACK_SIZE for some other reason.

sdbondi: "This function will only error if the new stack size exceeds the maximum allowed stack size of 256 (previous value)". - to exceed 256, the new stack size must be 257 or greater. Was a pedantic change so no problems with changing it back :) will do so

sdbondi: Ah I'm finally with you, the MAX_STACK_SIZE should be 255 with the `>=` - damn off-by-one errors!

sdbondi:
We currently don't impose a byte limit on the script in consensus, but makes sense to do it in validation rather than a hard coded limit in tari_crypto. 👍

CjS77:
> We currently don't impose a byte limit on the script in consensus, but makes sense to do it in validation rather than a hard coded limit in tari_crypto. 👍

Agree. I was editing my last comment when you posted this.

CjS77 (approved):

PR #66, opened 2021-11-19
[wip] refactor!: separate pub keys and signatures into compressed and noncompressed traits

mikethetike:
Compressing and decompressing Ristretto points can take a long time and was a large portion of the CPU time in the Tari project. 

This PR splits the traits into two for each of PublicKey, Signature, and CommitmentSignature.

The original trait no longer implements `ByteArray`, `Serialize` or `Deserialize` and must be compressed before serializing. 

Probably the biggest gainer in the `tari-crypto` crate itself is that points are no longer compressed when `Adding` or `Multiplying`.

Below is one of the benchmarks showing the new time for verifying signatures (red) against the old (blue) 
image

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.11.2"
 edition = "2018"
 
 [dependencies]
-tari_utilities = "^0.3"
+tari_utilities = { git = "https://github.com/tari-project/tari_utilities", branch="main"}
 base64 = "0.10.1"
 digest = "0.9.0"
 rand = { version = "0.8", default-features = false }
--- a/benches/range_proof.rs
+++ b/benches/range_proof.rs
@@ -30,11 +30,20 @@ use tari_crypto::{
     ristretto::{
         dalek_range_proof::DalekRangeProofService,
         pedersen::{PedersenCommitment, PedersenCommitmentFactory},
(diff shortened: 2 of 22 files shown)


CjS77 (commented):
This looks like multiple PRs in one. 

There are some changes that are great, and unrelated to the compress/decompress performance issue; which IMO we can merge in without any fuss.

I have some concerns about the approach taken to reduce superfluous compress/decompress calls (which was ignored until now as a case of premature optimisation - but it looks like there's concrete data to support looking at this now).

Firstly, based on a quick scan, it looks like it breaks the current API while adding leaky abstractions. There are several projects that use this crate that are not related to Tari which would prevent them from upgrading without doing a big refactor.
Secondly, it looks like there are cases where performance would be worse under these changes, but I'd like a chance to take some time to confirm this and write some test cases.

BTW I think we can implement the performance improvements without any of the downsides at all.

mikethetike:
I thought I had deleted these, but here is the report on the benchmarks in tari_crypto. The bulletproof library takes in a compressed point already, so no change there, and Generate Secret key is also unaffected by these changes

[criterion-development-compared.zip](https://github.com/tari-project/tari-crypto/files/7569975/criterion-development-compared.zip)

PR #69, opened 2022-01-05
perf: lazily compress RistrettoPublicKey

sdbondi:
Implement lazy compression for RistrettoPublicKey using thread-safe [`OnceCell`](https://docs.rs/once_cell/latest/once_cell/index.html) (currently unstable in std) 

image

Red is on current main branch, blue are the changes in the PR

- Minor optimisation for lexicographical public key ordering: using reference to the underlying compressed array rather than copying the array onto the stack and then creating a reference

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,21 +12,23 @@ edition = "2018"
 
 [dependencies]
 tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", branch = "main" }
+
 base64 = "0.10.1"
-digest = "0.9.0"
-rand = { version = "0.8", default-features = false }
-getrandom = { version = "0.2.3", default-features = false, optional = true }
-curve25519-dalek = { package = "curve25519-dalek-ng", version = "4.1", default-features = false, features = ["u64_backend", "serde", "alloc"] }
-bulletproofs = { version = "4.1", package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", branch = "main" } # todo: update when 4.1 pushed to crates.io
-merlin = { version = "3", default-features = false }
-sha2 = "0.9.5"
-sha3 = "0.9"
-thiserror = "1.0.20"
 blake2 = "0.9.1"
+bulletproofs = { version = "4.1", package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", branch = "main" } # todo: update when 4.1 pushed to crates.io
+curve25519-dalek = { package = "curve25519-dalek-ng", version = "4.1", default-features = false, features = ["u64_backend", "serde", "alloc"] }
+digest = "0.9.0"
+getrandom = { version = "0.2.3", default-features = false, optional = true }
+lazy_static = "1.3.0"
+merlin = { version = "3", default-features = false }
+once_cell = "1.8.0"
+rand = { version = "0.8", default-features = false }
 rmp-serde = "0.13.7"
 serde = "1.0.89"
 serde_json = "1.0"
-lazy_static = "1.3.0"
+sha2 = "0.9.5"
+sha3 = "0.9"
+thiserror = "1.0.20"
 wasm-bindgen = { version = "^0.2", features = ["serde-serialize"], optional = true }
 zeroize = { version = "1.3.0", default-features = false, features = ["alloc"] }
 
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,10 +34,14 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --all-features
+          # TODO: re-add all features once https://github.com/rust-lang/packed_simd/pull/341 is used by dalek
+          # args: --all-features
+          args: --features wasm --features ffi --features no_cc
       - name: test/release all features
         if: ${{ matrix.rust != 'stable' }}
         uses: actions-rs/cargo@v1
         with:
           command: test
-          args: --release --all-features
+          # TODO: re-add all features once https://github.com/rust-lang/packed_simd/pull/341 is used by dalek
+          # args: --release --all-features
+          args: --release --features wasm --features ffi --features no_cc
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -36,7 +36,7 @@ jobs:
           command: test
           # TODO: re-add all features once https://github.com/rust-lang/packed_simd/pull/341 is used by dalek
           # args: --all-features
-          args: --features wasm --features ffi --features no_cc
+          args: --features wasm --features ffi
       - name: test/release all features
         if: ${{ matrix.rust != 'stable' }}
         uses: actions-rs/cargo@v1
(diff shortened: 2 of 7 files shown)


sdbondi:
[criterion.tar.gz](https://github.com/tari-project/tari-crypto/files/7813167/criterion.tar.gz)

Cargo.toml
```diff
@@ -13,21 +13,22 @@ edition = "2018"
 [dependencies]
 tari_utilities = "^0.3"
 base64 = "0.10.1"
-digest = "0.9.0"
-rand = { version = "0.8", default-features = false }
-getrandom = { version = "0.2.3", default-features = false, optional = true }
+blake2 = "0.9.1"
```
sdbondi, line 16: OCD sort

CjS77 (approved):
This LGTM, but some thoughts for further work:
* I commented on a previous PR that I really don't like the `CompressedRistretto` struct. It's an ugly leaky abstraction that we can do away with once we have full lazy evaluation. Clients of this library should not have to think "do I need a compressed or uncompressed key here"? Or "wtf is a compressed key? I just want to multiple `k` by `G`".
* Secondly, I think we get additional benefit from making both the compressed and uncompressed private fields lazy. From my take, most of the perf gains come when pubkeys are simply passed along (in compressed form) and never do an ECC operation. Thus they never need to be decompressed.
* Knowing when to use compressed vs decompressed is actually obvious to the library. Compressed keys are always coming from I/O (or Serialization / Deserialization). "Real math" always uses the uncompressed form. So we can drop `CompressedRistretto` and let the `RistrettoPublicKey` take that cognitive load of users of the library.

sdbondi:
Thanks for the review 
> Secondly, I think we get additional benefit from making both the compressed and uncompressed private fields lazy. From my take, most of the perf gains come when pubkeys are simply passed along (in compressed form) and never do an ECC operation. Thus they never need to be decompressed.

Agree, the snafu is that we don't know that the compressed data is valid until we decompress. Only the caller will know if they can "trust" the data (i.e already been checked). If we decompress lazily (I implemented this too) we practically have no choice but to panic in `Add` `Mul` etc if the data is invalid. I had an idea to allow the caller the option to say "I trust this data" to avoid decompression (`from_bytes_unchecked`) but left it out for now due to the danger of misuse.

mikethetike:
I agree with your points on abstraction and I'm happy to close the other PR.

I'll let @sdbondi explain the difficulty in making the decompress step lazy. We haven't abandoned that idea, this PR is a step forward to making them both lazy. Both @sdbondi and I have spent a few days trying out RefCells and UnsafeCells running into many problems.

CjS77:
Breaking changes need a "major" version upgrade (0.12.0)

sdbondi:
This isn't a breaking change AFAICS (de)serialization is unaffected and there are no public interface changes

CjS77:
Oh, were all the `compress()` calls within the crate? Ok, np then

CjS77:
> Thanks for the review
> 
> > Secondly, I think we get additional benefit from making both the compressed and uncompressed private fields lazy. From my take, most of the perf gains come when pubkeys are simply passed along (in compressed form) and never do an ECC operation. Thus they never need to be decompressed.
> 
> Agree, the snafu is that we don't know that the compressed data is valid until we decompress. Only the caller will know if they can "trust" the data (i.e already been checked). If we decompress lazily (I implemented this too) we practically have no choice but to panic in `Add` `Mul` etc if the data is invalid. I had an idea to allow the caller the option to say "I trust this data" to avoid decompression (`from_bytes_unchecked`) but left it out for now due to the danger of misuse.

hmm.. good point.
It's important that `RistrettoPublicKey` instances are always valid and fail early.

So it looks like we have no choice but to keep `CompressedRistrettoKey` in some form. Perhaps we should change the name to reflect its intended use case and the fact that it might not represent a valid pubkey.

CjS77:
We could think about introducing an additional trait that captured this. Something like

```
/// A serialized, often more compact representation, of a public key. You cannot do math operations with instances,
/// nor is the key guaranteed to be valid, or actually on the curve. You can use `to_public_key` to validate that
/// the key is correct and if necessary, perform any calculations to allow mathematical operations. 
pub trait SerializedPublicKey {
    type P: PublicKey;
    type E: Error;
    
    /// Try convert a serialized public key to its canonical form. 
    fn to_public_key(self) -> Result<P, E>;
    
}
```

This will help guide users of the library (which are not just the tari project) use only the public API from this crate rather than reaching down into Ristretto and locking themselves into that concrete implementation.

For Ristretto, we can then have a newtype over the compressed point and impl this trait without incurring any performance overhead.

sdbondi:
Yeah that LGTM, esp approach to avoid the danger of potentially invalid data by using the type system and client code can be optimised slowly to use it 👍

CjS77:
I know performance is important obviously, but I also always want to achieve these two goals in this crate:
- An intuitive, elegant, expressive API that makes crypto code "look like the math" as much as possible.
  This really helps in places like RustPen and TLU where tutorials using crypto are very clean and don't have any unnessecary boilerplate (libsecp256k1's API is really awkward in this regard).
- Being able to change the underlying curve (assuming we eventually support other curves, like ed25519 and secp256k1) without changing client code beyond the actual type used.

delta1:
Some info regarding the build failure: dtolnay/proc-macro2#307 

We'll either have to upgrade the nightly toolchain, or pin this transitive dependency to 1.0.32 (although since the lockfile is not committed in libs I'm not sure that's possible?) 

Compiling with stable toolchain does still work. 
Just tested with nightly-2022-01-03 which does work, and did also work on tari-project/tari repo a few days ago when I tried that (minor tweaks)

sdbondi:
I looked at this for a few minutes and even when locking the dependency `proc-macro2 = {version="=1.0.32", optional = true}` ran into other issues (no function or associated item named `from_str_unchecked` found for struct `proc_macro2::Literal` in the current scope).

delta1:
> I looked at this for a few minutes and even when locking the dependency `proc-macro2 = {version="=1.0.32", optional = true}` ran into other issues (no function or associated item named `from_str_unchecked` found for struct `proc_macro2::Literal` in the current scope).

Yeah I see the same. Probably worth bumping the toolchain instead of trying to play old dependency whack-a-mole.

sdbondi:
Let's carry on discussions for solutions in #74

Cargo.toml
```diff
@@ -47,7 +47,7 @@ simd = ["simd_backend"]
 avx2 = ["simd_backend"] # deprecated alias for simd_backend
 wasm = ["wasm-bindgen", "getrandom/js"]
 ffi = []
-no_cc_nightly = ["clear_on_drop/nightly"]
+no_cc_nightly = []
```
CjS77, line 50: This feature exists to compile clear_on_drop for envs like wasm which struggle when needing a C compiler at times.
Was the  `nightly` feature removed from the `clear_on_drop` crate?

delta1: No I think it’s failing to compile with that feature on this nightly 

https://github.com/tari-project/tari-crypto/runs/4854291267?check_suite_focus=true

sdbondi: yup, llvm_asm! support was dropped in rust and this feature uses it. We want to replace clear_on_drop with `zeroize` since clear_on_drop doesn't seem actively maintained.

delta1: Zeroize is also used in the underlying dalek crate

sdbondi: #75 has been merged

mikethetike (approved):

PR #74, opened 2022-01-07
chore!: update rust toolchain (proc_macro_is_available feature)

sdbondi:
Checking if CI works


delta1:
That test failure is very strange

sdbondi:
Yeah very strange, returns an error code for utf8 conversion for the string "msg" - works fine in locally on my mac. Trying on Linux and will see if I can reproduce there.

sdbondi:
Get's even stranger, but on Linux in release the string must be null terminated `"msg\0" - this is only required on release mode on linux.

.github/workflows/clippy-check.yml
```diff
@@ -8,6 +8,8 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           components: clippy, rustfmt
+          toolchain: nightly-2022-01-17
```
mikethetike, line 11: Don't think this is needed

mikethetike (approved):

PR #76, opened 2022-03-01
chore: point tari_utilities to the git repo

philipr-za:
We want to be able to use the latest tari_utilities in the main Tari repo rather than having to push new versions to crates but `tari_crypto` re-exports the crates version.

So updating `tari-crypto` to point at the `tari_utilities` git repo too for faster iteration.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,6 @@ jobs:
         rust:
           - stable
           - nightly-2022-01-17
-          - nightly
     steps:
       - name: checkout
         uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


CjS77 (dismissed):

philipr-za:
Hmm I am not sure what is up with the CI testing here? Locally with the toolchain specified in the repo is passes

philipr-za:
@CjS77 we currently have the latest nightly in the test matrix as well as the toolchain specified in `rust-toolchain`. Any objections to removing that version from the tests and just testing with the `rust-toolchain` version?

delta1 (approved):

philipr-za:
@CjS77 @mikethetike I see that the CI steps that are stuck here are CircleCI steps but it looks like the circle CI config has been removed from the repo a while back? Should these steps maybe be removed?

CjS77:
Right. I'll do that now

PR #81, opened 2022-03-10
chore!: remove tari script

### LINKED ISSUE #70: Move TariScript out of tari-crypto crate
Developing TariScript in this crate was useful since it forced a clean separation from any dependencies in the Tari blockchain project, but it really doesn't belong here.

Migrating it to its own crate in tari-project/tari is probably the best move.

sdbondi:
Remove Tari script from tari crypto (closes #70)
Ref https://github.com/tari-project/tari/pull/3909


sdbondi:
Since we currently reference tari_crypto using a git ref, DNM until https://github.com/tari-project/tari/pull/3909 is merged

mikethetike (approved):

PR #86, opened 2022-03-14
ci: add wasm build artifacts and test run

delta1:
- Adds a new github action to run on pull requests and pushes to the main branch.
- Checks that wasm compiles and tests successfully 
- If the action is triggered on push to `main` then the wasm artifacts are uploaded as well


sdbondi (approved):
👍

mikethetike (approved):

PR #88, opened 2022-03-16
feat: use standard-version

sdbondi:
Adds script using `standard-version` to bump versions in Cargo.toml and update changelog

## Usage

`./scripts/bump-version`

**prereleases**

`./scripts/bump-version --prerelease rc`

https://github.com/conventional-changelog/standard-version#cli-usage

### DIFF
--- a/scripts/bump-version
+++ b/scripts/bump-version
@@ -1,0 +1,8 @@
+#!/usr/bin/env bash
+
+set -e
+
+pushd `git rev-parse --show-toplevel` > /dev/null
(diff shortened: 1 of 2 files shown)


sdbondi:
That would be great, but unfortunately not. At least there is no need to run `npm install`/ gitignore `node_modules` in the repo, it's a (optional) convenience tool and having nodejs installed isn't uncommon.

sdbondi:
Perhaps we can get rid of the script since all it does is run `npx standard-version` anyway.

sdbondi:
Closing in favour of #89

PR #97, opened 2022-05-20
chore: split range proof trait

hansieodendaal:
Split rewindable range proofs from the range proof trait into its own trait in preparation for adding Bulletproofs+, as not all rangeproofs are rewindable.

### DIFF
--- a/src/rewindable_range_proof.rs
+++ b/src/rewindable_range_proof.rs
@@ -1,0 +1,107 @@
+// Copyright 2019. The Tari Project
+//
+// Redistribution and use in source and binary forms, with or without modification, are permitted provided that the
+// following conditions are met:
+//
+// 1. Redistributions of source code must retain the above copyright notice, this list of conditions and the following
+// disclaimer.
+//
+// 2. Redistributions in binary form must reproduce the above copyright notice, this list of conditions and the
+// following disclaimer in the documentation and/or other materials provided with the distribution.
+//
+// 3. Neither the name of the copyright holder nor the names of its contributors may be used to endorse or promote
+// products derived from this software without specific prior written permission.
+//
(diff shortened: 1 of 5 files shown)


src/ristretto/dalek_range_proof.rs
```diff
@@ -103,6 +99,12 @@ impl RangeProofService for DalekRangeProofService {
     fn range(&self) -> usize {
         self.range
     }
+}
+
+impl RewindableRangeProofService for DalekRangeProofService {
+    type K = RistrettoSecretKey;
+    type P = Vec<u8>;
```
sdbondi, line 106: Think `Proof` is a better name
```suggestion
    type Proof = Vec<u8>;
```

sdbondi (commented):
LGTM - just one comment

src/rewindable_range_proof.rs
```diff
@@ -0,0 +1,107 @@
...
+    /// Construct a rangeproof with the ability to rewind it. Requires two rewind keys and a 19-byte message to be
+    /// included in the range proof. The proof can contain 23 bytes but 4 bytes are used to confirm that a rewind
+    /// was performed correctly
+    fn construct_proof_with_rewind_key(
+        &self,
+        key: &Self::K,
+        value: u64,
+        rewind_key: &Self::K,
+        rewind_blinding_key: &Self::K,
+        proof_message: &[u8; REWIND_USER_MESSAGE_LENGTH],
```
CjS77, line 45: Fine for this PR, since it was like this, but ideally, this should be general and not locked to 21 bytes for every impl of the trait. `&[u8]` will work here.

src/ristretto/dalek_range_proof.rs
```diff
@@ -66,7 +62,7 @@ impl DalekRangeProofService {
             B_blinding: base.G,
             B: base.H,
         };
-        let bp_gens = BulletproofGens::new(64, 1);
+        let bp_gens = BulletproofGens::new(range, 1);
```
CjS77, line 65: Good catch

CjS77 (approved):
Nice LTGM

PR #98, opened 2022-05-25
chore: swap out curve25519-dalek-ng for curve25519-dalek

hansieodendaal:
Swapped out curve25519-dalek-ng for curve25519-dalek in preparation for bulletproofs-plus


CjS77 (approved):
LGTM - no detailed testing required

PR #99, opened 2022-05-26
feat: add extended ristretto commitment factory and pedersen generators

hansieodendaal:
Added extended ristretto commitment factory for Pedersen commitments making use of the extended Pedersen generators in the Tari Bulletproofs+ library.

Notes:
- `RISTRETTO_NUMS_POINTS` have been updated by sequentially hashing the domain separated Ristretto255 generator point instead of sequentially hashing the generator point only. Domain separated hashing is less likely to result in unintended collisions or reuse elsewhere.
- An `ExtendedHomomorphicCommitmentFactory` trait has been added
- `ristretto/pedersen` has been changed into a module that defines `pedersen/commitment_factory` and `pedersen/extended_commitment_factory`
- `PedersenCommitmentFactory` and `ExtendedPedersenCommitmentFactory` use the same base `G` and `H` generator points so that each factory produces the same commitments for their default implementation.
- For `ExtendedPedersenCommitmentFactory`:
  - Homomorphism with public key only holds for extended commitments with `ExtensionDegree::Zero`
  - ~'Ord' and 'PartialOrd' are undefined for extended commitments other than `ExtensionDegree::Zero`~
- _**Update**: The common interface trait `HomomorphicCommitmentFactory` has also been implemented for `ExtendedPedersenCommitmentFactory` so that its default invocation is compatible with all methods of `PedersenCommitmentFactory`._
- 
_The next PR will include the `ExtendedRangeProofService` trait and `BulletProofsPlusService`._

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
-bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", tag = "v0.0.1" }
+bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", tag = "v0.0.3" }
 curve25519-dalek = { package = "curve25519-dalek", version = "4.0.0-pre.2", default-features = false, features = ["serde", "alloc"] }
 digest = "0.9.0"
 getrandom = { version = "0.2.3", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
-bulletproofs_plus = { path = "../bulletproofs-plus", package = "tari_bulletproofs_plus"}
+bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus"}
 curve25519-dalek = { package = "curve25519-dalek", version = "4.0.0-pre.2", default-features = false, features = ["serde", "alloc"] }
 digest = "0.9.0"
 getrandom = { version = "0.2.3", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
-bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus"}
+bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", tag = "v0.0.1" }
 curve25519-dalek = { package = "curve25519-dalek", version = "4.0.0-pre.2", default-features = false, features = ["serde", "alloc"] }
 digest = "0.9.0"
 getrandom = { version = "0.2.3", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
-bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", tag = "v0.0.3" }
 curve25519-dalek = { package = "curve25519-dalek", version = "4.0.0-pre.2", default-features = false, features = ["serde", "alloc"] }
 digest = "0.9.0"
 getrandom = { version = "0.2.3", default-features = false, optional = true }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,6 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
+bulletproofs_plus = { path = "../bulletproofs-plus", package = "tari_bulletproofs_plus"}
 curve25519-dalek = { package = "curve25519-dalek", version = "4.0.0-pre.2", default-features = false, features = ["serde", "alloc"] }
 digest = "0.9.0"
 getrandom = { version = "0.2.3", default-features = false, optional = true }
--- a/src/ristretto/constants.rs
+++ b/src/ristretto/constants.rs
@@ -21,9 +21,11 @@
 // USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 
 use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
-/// These points on the Ristretto curve have been created by sequentially hashing a domain separated Generator point
-/// with SHA512 and using the byte string representation of the hash as input into the `from_uniform_bytes` constructor
-/// in [RistrettoPoint](Struct.RistrettoPoint.html). This process is validated with the `check_nums_points` test below.
+
+/// These points on the Ristretto curve have been created by hashing domain separation labels with SHA512 and converting
+/// the hash output to a Ristretto generator point by using the byte string representation of the hash as input into the
+/// `from_uniform_bytes` constructor in [RistrettoPoint](Struct.RistrettoPoint.html). This process is validated with the
+/// `check_nums_points` test below.
 pub const RISTRETTO_NUMS_POINTS_COMPRESSED: [CompressedRistretto; 10] = [
     CompressedRistretto([
         206, 56, 152, 65, 192, 200, 105, 138, 185, 91, 112, 36, 42, 238, 166, 72, 64, 177, 234, 197, 246, 68, 183, 208,
@@ -80,7 +82,10 @@ lazy_static! {
 
 #[cfg(test)]
 mod test {
-    use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint};
+    use curve25519_dalek::{
+        constants::{RISTRETTO_BASEPOINT_COMPRESSED, RISTRETTO_BASEPOINT_POINT},
+        ristretto::{CompressedRistretto, RistrettoPoint},
+    };
     use sha2::{Digest, Sha512};
 
     use crate::ristretto::constants::{RISTRETTO_NUMS_POINTS, RISTRETTO_NUMS_POINTS_COMPRESSED};
@@ -93,8 +98,8 @@ mod test {
         let mut compressed_points = Vec::with_capacity(n);
         let mut a: [u8; 64] = [0; 64];
         for i in 0..n {
-            let mut data = b"TARI CRYPTO NUMS BASEPOINT LABEL - ".to_vec();
-            data.append(&mut i.to_le_bytes().to_vec());
+            let mut data = b"TARI CRYPTO NUMS BASEPOINT LABEL - ".to_vec(); // Domain label
+            data.append(&mut i.to_le_bytes().to_vec()); // Append domain separated label counter
             let hashed_v = Sha512::digest(&*data);
             a.copy_from_slice(&hashed_v);
             let next_val = RistrettoPoint::from_uniform_bytes(&a);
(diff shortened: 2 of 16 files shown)


Cargo.toml
```diff
@@ -16,6 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
+bulletproofs_plus = { path = "../bulletproofs-plus", package = "tari_bulletproofs_plus"}
```
delta1, line 19: You'll need to point this to github I think

hansieodendaal: Yeah, looking at that now, forgot about it

src/ristretto/constants.rs
```diff
@@ -92,15 +88,18 @@ mod test {
...
         let mut val = RISTRETTO_BASEPOINT_POINT.compress().to_bytes();
         let mut points = Vec::with_capacity(n);
         let mut compressed_points = Vec::with_capacity(n);
         let mut a: [u8; 64] = [0; 64];
-        for _ in 0..n {
-            let hashed_v = Sha512::digest(&val[..]);
+        for i in 0..n {
+            let mut data = b"TARI CRYPTO - ".to_vec();
+            data.append(&mut i.to_le_bytes().to_vec());
+            data.append(&mut val.to_vec());
```
AaronFeickert, line 101: Is there a particular reason to bind the base point to these generators? They're still NUMS if you simply use an indexed hash. I also recommend using a more descriptive domain separator that makes the intent of the generators more clear.

Depending on efficiency, you could also use `Blake2b`, which has specific safe functionality for domain separation (the `persona` parameter) and index-type functionality (the `salt` parameter, or simply passing the index as fixed-encoding input).

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    fn zero(&self) -> PedersenCommitment {
+        let extension_degree = self.0.extension_degree as usize;
+        let zero = vec![Scalar::zero(); extension_degree + 1];
+        let mut points = Vec::with_capacity(extension_degree + 1);
+        points.push(self.0.h_base);
+        points.append(&mut self.0.g_base_vec.clone());
+        let c = RistrettoPoint::multiscalar_mul(&zero, &self.0.g_base_vec);
+        HomomorphicCommitment(RistrettoPublicKey::new_from_pk(c))
+    }
```
AaronFeickert, line 90: This must always return the group identity, so there's no need to compute this multiscalar multiplication.

AaronFeickert (changes requested):

AaronFeickert (changes requested):
It may be useful to add a test that asserts the NUMS points are unique and all non-zero, to catch the case where the hashing implementation has failed catastrophically. This isn't sufficient to guarantee the generators are sufficiently independent, but the failure of such a test would immediately imply commitments are trivially non-binding.

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,415 @@
...
+            let mut hasher = DefaultHasher::new();
+            c1.hash(&mut hasher);
+            let c2 = c1.clone();
+            let c3 = factory.commit_value(&k, 2049).unwrap();
+
+            // Test 'clone`
+            assert_eq!(c1, c2);
+            assert_ne!(c2, c3);
+
+            // Note! 'Ord' and 'PartialOrd' are undefined for extended commitments with extension degree other than Zero
```
AaronFeickert, line 346: Why undefined? Point ordering is lexicographic on canonical byte representations, and makes sense for any group element. (That being said, ordering has no algebraic significance whatsoever...)

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,415 @@
...
+                        format!("{:?}", c1),
+                        "HomomorphicCommitment(9801c7785217e0c973e9b85508c6eebcb74b257a6f825630e17282b81b7fcd78)"
+                    );
+                    // Test hashing
+                    let result = format!("{:x}", hasher.finish());
+                    assert_eq!(&result, "1ef11e8d243c886e");
+                    // test 'Ord' and 'PartialOrd'
+                    assert!(c2 > c3);
+                    assert!(c3 < c2);
+                    assert!(matches!(c2.cmp(&c3), std::cmp::Ordering::Greater));
```
AaronFeickert, line 360: In case there's any misunderstanding, point ordering is lexicographic on canonical byte representations, and has no algebraic significance of any kind (in the group structure). Performing algebraic operations on points (in this example, committing to a larger or smaller value with the same mask) is not guaranteed or expected to lead to any specific ordering of them. Ordering is useful for operations like sorting for hashing purposes, where a given set of points needs a canonical ordering.

src/ristretto/pedersen/extended_commitment_factory.rs
hansieodendaal: I think what I should have said is that for this specific test _"When manipulating v and k_i, ordering of the commitment result is unpredictable for extended commitments with extension degree other than Zero, but certainly repeatable"_. Anyways, I fixed the test to be robust with changing extension degrees

Cargo.toml
```diff
@@ -16,6 +16,7 @@ tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", t
 base64 = "0.10.1"
 blake2 = "0.9.1"
 bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", tag = "v4.2.0" }
+bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus"}
```
mikethetike, line 19: Need to reference a tag

src/ristretto/constants.rs
hansieodendaal: Updated as suggested, although I did not change the hash function as hashing performance is not an issue for the test case that generates the test vector for the NUMS constants.

src/ristretto/pedersen/extended_commitment_factory.rs
hansieodendaal: Updated and added a specific test case for it

src/ristretto/pedersen/commitment_factory.rs
```diff
@@ -275,31 +249,43 @@ mod test {
...
-    }
+        assert_eq!(&result, "699d38210741194e");
 
-    #[test]
-    fn default_value() {
-        let c = PedersenCommitment::default();
-        assert_eq!(c, PedersenCommitment::from_public_key(&RistrettoPublicKey::default()));
+        // Test 'Ord' and 'PartialOrd' implementations
+        let mut values = (value - 100..value - 1).collect::<Vec<_>>();
+        values.extend((value + 1..value + 100).collect::<Vec<_>>());
```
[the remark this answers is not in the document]
CjS77: I'd just keep the test simple and choose one c2 value, check its lexicographic value out of band to check if it's bigger than c1 and do the comparison test.
then find a value that's smaller and check

AaronFeickert (dismissed):

src/ristretto/pedersen/mod.rs
```diff
@@ -0,0 +1,165 @@
...
+                PedersenCommitment,
+                RISTRETTO_PEDERSEN_G,
+                RISTRETTO_PEDERSEN_H,
+            },
+            RistrettoPublicKey,
+            RistrettoSecretKey,
+        },
+    };
+
+    static EXTENSION_DEGREE: [ExtensionDegree; 6] = [
```
CjS77, line 86: DRY - Move the test that uses this to `extended_commitment_factory.rs` and then you don't have to duplicate this code.

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,451 @@
...
+    /// Create a new Extended Pedersen Ristretto Commitment factory for the required extension degree using
+    /// pre-calculated compressed constants - we only hold references to the static generator points.
+    pub fn new_with_extension_degree(extension_degree: ExtensionDegree) -> Result<Self, RangeProofError> {
+        if extension_degree as usize > RISTRETTO_NUMS_POINTS.len() ||
+            extension_degree as usize > RISTRETTO_NUMS_POINTS_COMPRESSED.len()
+        {
+            return Err(RangeProofError::ExtensionDegree(
+                "Not enough Ristretto NUMS points to construct the extended commitment factory".to_string(),
+            ));
+        }
```
CjS77, line 60: I guess we can generalise this in future. But fine for now.

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,451 @@
...
+        let mut g_base_vec = Vec::with_capacity(extension_degree as usize);
+        g_base_vec.push(RISTRETTO_PEDERSEN_G);
+        let mut g_base_compressed_vec = Vec::with_capacity(extension_degree as usize);
+        g_base_compressed_vec.push(RISTRETTO_PEDERSEN_G_COMPRESSED);
+        for i in 1..extension_degree as usize {
+            g_base_vec.push(RISTRETTO_NUMS_POINTS[i]);
+            g_base_compressed_vec.push(RISTRETTO_NUMS_POINTS_COMPRESSED[i]);
+        }
```
CjS77, line 68: doesn't `RISTRETTO_NUMS_POINTS[..extension_degree as usize].to_vec()` do all of this in one line? Same for compressed

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,451 @@
...
+
+impl ExtendedHomomorphicCommitmentFactory for ExtendedPedersenCommitmentFactory {
+    type P = RistrettoPublicKey;
+
+    fn commit(
+        &self,
+        k_i: &[RistrettoSecretKey],
+        v: &RistrettoSecretKey,
+    ) -> Result<PedersenCommitment, RangeProofError> {
+        let k_i: Vec<Scalar> = k_i.to_vec().iter().map(|k| k.0).collect();
```
CjS77, line 95: Unnecessary

CjS77 (commented):
This is a partial overview review. 

Everything is reviewed except `extended_commitment_factory.rs` which requires a deeper dive, so I'll focus on that in a 2nd pass.

Everything else is fine, except some semantic and nit-picky stuff. See comments

src/ristretto/pedersen/extended_commitment_factory.rs
CjS77: It's actually a bit more complicated since G is not point 0, but it's still nice and compact:

```rust
    let g_base_vec: std::iter::once(&RISTRETTO_PEDERSEN_G).chain(RISTRETTO_NUMS_POINTS[1..n].iter()).cloned().collect();
    let g_base_compressed_vec = std::iter::once(&RISTRETTO_PEDERSEN_G_COMPRESSED).chain(RISTRETTO_NUMS_POINTS_COMPRESSED[1..n].iter()).cloned().collect();
```

CjS77 (commented):
Reviewed `extended_commitment_factory.rs`. 

No major issues; the copies in `commit` are likely detrimental to performance, but I didn't run any benches to determine the extent of it.

src/ristretto/pedersen/commitment_factory.rs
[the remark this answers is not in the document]
hansieodendaal: Improved the efficiency of the test

src/ristretto/pedersen/mod.rs
hansieodendaal: I have added this test at the module level as it logically belongs there, but move the code as suggested in favour of DRY.

src/ristretto/pedersen/extended_commitment_factory.rs
hansieodendaal: Logged the issue [here](https://github.com/tari-project/bulletproofs-plus/issues/9).

src/ristretto/constants.rs
```diff
@@ -79,47 +80,58 @@ lazy_static! {
...
-        let mut val = RISTRETTO_BASEPOINT_POINT.compress().to_bytes();
         let mut points = Vec::with_capacity(n);
         let mut compressed_points = Vec::with_capacity(n);
         let mut a: [u8; 64] = [0; 64];
-        for _ in 0..n {
-            let hashed_v = Sha512::digest(&val[..]);
+        for i in 0..n {
+            let mut data = b"TARI CRYPTO NUMS BASEPOINT LABEL - ".to_vec(); // Domain label
+            data.append(&mut i.to_le_bytes().to_vec()); // Append domain separated label counter
+            let hashed_v = Sha512::digest(&*data);
```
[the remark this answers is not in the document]
hansieodendaal: Logged issue [here](https://github.com/tari-project/tari-crypto/issues/100).

src/ristretto/pedersen/extended_commitment_factory.rs
hansieodendaal: Logged issue [here](https://github.com/tari-project/tari-crypto/issues/101).

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -0,0 +1,697 @@
...
+/// [RistrettoPoints](curve25519_dalek::ristretto::RistrettoPoints).
+/// Notes:
+///  - Homomorphism with public key only holds for extended commitments with `ExtensionDegree::Zero`
+#[derive(Debug, PartialEq, Clone)]
+pub struct ExtendedPedersenCommitmentFactory(pub(crate) PedersenGens<RistrettoPoint>);
+
+impl ExtendedPedersenCommitmentFactory {
+    /// Create a new Extended Pedersen Ristretto Commitment factory for the required extension degree using
+    /// pre-calculated compressed constants - we only hold references to the static generator points.
+    pub fn new_with_extension_degree(extension_degree: ExtensionDegree) -> Result<Self, CommitmentError> {
```
stringhandler, line 53: I'm not too fond of the use of ExtensionDegree here. The type is in the `bulletproofs_plus` crate, meaning that you can't use this crate without Bulletproofs, for example, if it were hidden behind a feature flag. `PedersonGens` is perhaps in the wrong crate, or perhaps we should use a field in the `ExtendedPedersonCommitmentFactory` `gens: Vec<RistrettoPoint>`

stringhandler (changes requested):
I like most of it, but the abstraction is not 100% correct.

CjS77 (dismissed):
LGTM

src/ristretto/pedersen/extended_commitment_factory.rs
hansieodendaal: As we are dependent on 
```
use bulletproofs_plus::{generators::pedersen_gens::ExtensionDegree, PedersenGens};
```
here and if we need to remove the dependency I think the most elegant solution would be to completely move `generators` out of `bulletproofs_plus` into its own crate.

If that solution is acceptable I propose we deal with that as a separate series of PRs.

CjS77: Good point. Would it be better to rewrite `ExtensionDegree` as a newtype around u64 rather than an enum, since the code in bp+ always converts it to an int anyway. 
You could then add an adapter type in tari_crypto to resolve the dependency issue.

stringhandler (approved):

PR #106, opened 2022-06-22
feat: hashing api

CjS77:
Naive hashing of data for various applications (key derivation,
signature challenges, hmacs, group generators, etc.) can have subtle
security implications if not done carefully. The hashing API strives to
adopt the implicit use of best practices for these applications using an
API that is:
* super ergonomic and stays largely out of the way.
* leverages the Rust type system to achieve zero-cost abstractions and
  guarantees that best practices are being followed.
* Performing these best practices in tari-crypto, and letting clients
  focus on their applications.

The API makes heavy use of traits (incl marker traits) and generics to
keep the API as flexible and broadly applicable as possible.

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -150,7 +150,7 @@ impl AsRef<[u8]> for DomainSeparatedHash {
 /// assert_eq!(id.domain_separation_tag(), "com.tari.generic.v1.card_id");
 /// assert_eq!(
 ///     to_hex(id.as_ref()),
-///     "597a73459101ebc5b56545f98bf0c93b1631844b0ff6b13f0123d796c5159a16"
+///     "44fb39bfdd20c93ddf542e4b2d1f4b06448aa1fa2b9c4d138d8e7bbb19aa7c65"
 /// );
 /// ```
 ///
@@ -180,7 +180,7 @@ impl AsRef<[u8]> for DomainSeparatedHash {
 /// );
 /// assert_eq!(
 ///     to_hex(challenge.as_ref()),
-///     "f358c075dd820679905447f27a4cf771c2aca91d3985a3f90b54dd2f8d025c47"
+///     "6cd8efe7d7f1673ed8cfc0ac67d6979eb50afdbf276adf5221caabfbfd01da8c"
 /// );
 /// ```
 pub struct DomainSeparatedHasher<D, M> {
@@ -192,12 +192,14 @@ pub struct DomainSeparatedHasher<D, M> {
 impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
     pub fn new<S>(label: S) -> Self
     where S: AsRef<str> {
-        let inner = D::new().chain(M::domain_separation_tag(&label).as_bytes());
-        Self {
+        let inner = D::new();
+        let mut result = Self {
             inner,
             label: label.as_ref().to_string(),
             dst: PhantomData::<M>::default(),
-        }
+        };
+        result.update(M::domain_separation_tag(&label).as_bytes());
+        result
     }
 
     /// Adds the data to the digest function by first appending the length of the data in the byte array, and then
@@ -292,7 +294,7 @@ impl DomainSeparation for MacDomain {
 /// assert_eq!(mac.domain_separation_tag(), "com.tari.mac.v1.api.auth");
 /// assert_eq!(
 ///     to_hex(mac.as_ref()),
-///     "9d36cc40c9e4831efbdd1cd4305fe78bd014b30abd47ae752bf6c15da8644443"
+///     "796eb496b6672b1b7c4021e603d6b833121d35cd282a1555e3f9dd2eda5658b8"
 /// );
 /// ```
 pub struct Mac {
@@ -366,12 +368,12 @@ impl Deref for Mac {
 /// let key_1 = wallet_keys(&key, 1).unwrap();
 /// assert_eq!(
 ///     key_1.to_hex(),
-///     "be4b10a3f6669b04c840c34a36312fa6f54df858d1b8a4ea46a7a9e82be2670e"
+///     "b778b8b5041fbde6c78be5bafd6d62633824bf303c97736d7337b3f6f70c4e0b"
 /// );
 /// let key_64 = wallet_keys(&key, 64).unwrap();
 /// assert_eq!(
 ///     key_64.to_hex(),
-///     "5c3c8fda79509681caa267bfdad872b912307c0b2bc49a317de534839ee3c60d"
+///     "09e5204c93406ef3334ff5f7a4d5d84199ceb9119fafcb98928fa95e95f0ae05"
 /// );
 /// ```
 pub trait DerivedKeyDomain: DomainSeparation {
@@ -402,6 +404,7 @@ pub trait DerivedKeyDomain: DomainSeparation {
 #[cfg(test)]
 mod test {
     use blake2::Blake2b;
+    use digest::Digest;
     use tari_utilities::hex::{from_hex, to_hex};
 
     use crate::{
@@ -430,8 +433,28 @@ mod test {
         assert_eq!(hash.as_ref(), hash2.as_ref());
         assert_eq!(
             to_hex(hash.as_ref()),
-            "9ec75b89b25f4c63eb39c717867b4330b66b612cc7c35d698612c9caeda4b281"
+            "a8326620e305430a0b632a0a5e33c6c1124d7513b4bd84736faaa3a0b9ba557f"
         );
+    }
+
+    #[test]
+    fn deconstruction() {
+        // Illustrate exactly what gets hashed and how we try and avoid collisions
+        let hash = DomainSeparatedHasher::<Blake256, GenericHashDomain>::new("mytest")
+            .chain("rincewind")
+            .chain("hex")
+            .finalize();
+        assert_eq!(hash.domain_separation_tag(), "com.tari.generic.v1.mytest");
+        assert_eq!(hash.domain_separation_tag().len(), 26);
+        let expected = Blake256::new()
+            .chain(26usize.to_le_bytes())
+            .chain("com.tari.generic.v1.mytest".as_bytes())
+            .chain(9usize.to_le_bytes())
+            .chain("rincewind".as_bytes())
+            .chain(3usize.to_le_bytes())
+            .chain("hex".as_bytes())
+            .finalize();
+        assert_eq!(hash.as_ref(), expected.as_slice());
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


src/hashing.rs
```diff
@@ -0,0 +1,437 @@
...
+}
+
+pub struct Kdf<K = RistrettoSecretKey>
+where K: SecretKey
+{
+    derived_key: K,
+}
+
+impl<K: SecretKey> Kdf<K> {
+    /// Generate a secure private key from the data provided in `key`. This might be a password, or some other source of
```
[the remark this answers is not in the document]
AaronFeickert: As an example of its use, for output scanning we (will) use the result of an ECDH exchange, which is considered to be of sufficiently high entropy for this purpose.

src/hashing.rs
```diff
@@ -0,0 +1,437 @@
...
+    /// Returns the version number for the metadata tag
+    fn version() -> u8;
+    /// Returns the category label for the metadata tag. For example, `tari_hmac`
+    fn tag() -> &'static str;
+    /// The domain separation tag is defined as `H(version | tag | output_size)`, where the version and tag are
+    /// typically hard-coded into the implementing type, and the output size is the length of the hash generated by
+    /// the hashing algorithm `D`.
+    fn hash_domain_separation_tag<D: Digest>(mut hasher: D) -> D {
+        hasher.update(&[Self::version()]);
+        encode_with_size(Self::tag(), hasher).chain(D::output_size().to_le_bytes())
```
AaronFeickert, line 81: There may have been some misunderstanding around the intent for embedding an output size. The "sampling process" for a hash function should include the output size _required by the use case_, which may differ from the output size of the underlying digest. For example, a `SipHash` key is 32 bytes, but an underlying digest might output 64 bytes. Embedding the expected output size ensures that an implementer can't accidentally reuse a label but specify a different output size, which could produce keys with a subset relationship.

src/hashing.rs
```diff
@@ -0,0 +1,437 @@
...
+    ) -> Result<Self, HashingError> {
+        // The input key does not have enough entropy to securely generate a secret key
+        if key.as_ref().len() < D::output_size() {
+            return Err(HashingError::InsufficientEntropy);
+        }
+        let hasher = encode_with_size(key, D::new());
+        let hasher = M::hash_domain_separation_tag(hasher);
+        let result = encode_with_size(label, hasher).finalize();
+
+        let derived_key = K::from_bytes(result.as_slice())?;
```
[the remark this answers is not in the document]
CjS77: Yes, it will fail.

src/hashing.rs
CjS77: Now you actually specify the type you want as the result of the key derivation, so the code can do additional checks.

DenisKolodin:
LGTM. It designed well. If domain has to be more dynamic (not sure that is necessary here), but the trait can be replaced or extended with a struct:

```rust
#[derive(Clone)]
pub struct Domain {
    version: u8,
    domain: Cow<'static, str>,
}

impl Domain {
    pub const fn global(domain: &'static str, version: u8) -> Self {
        Self { version, domain: Cow::Borrowed(domain) }
    }

    pub fn local(domain: impl ToString, version: u8) -> Self {
        Self { version, domain: Cow::Owned(domain.to_string()) }
    }

    pub fn scope(&self, label: impl AsRef<str>) -> Scope {
        let fqdn = format!("{}.v{}.{}", self.version, self.domain, label.as_ref());
        Scope { fqdn }
    }
}

pub struct Scope {
    fqdn: String,
}

pub const MAC: Domain = Domain::global("com.tari.mac", 1);
pub const GENERIC: Domain = Domain::global("com.tari.generic", 1);

//...

let scope = MAC.scope("subdomain");
```

The `DomainSeparation` trait can be removed, or preserved and implemented for the `Domain` struct that allows to keep and use `DerivedKeyDomain` trait as well.

CjS77:
Nice suggestion. As you surmised, the goal is to have the domain separation tags be "almost" hardcoded, so the dynamism is less desirable in this case.

CjS77:
> > Nice suggestion. As you surmised, the goal is to have the domain separation tags be "almost" hardcoded, so the dynamism is less desirable in this case.
> 
> Even though there's a canonical form to the domain separators, their length variability (without length prepending) could still result in collisions, right?

That's the one place where length is not prepended. It's trivial to add.
I think some tests indicating _what is actually hashed_ once this is all unpacked may be useful too.

CjS77:
tag length is included in the hash now, and another test provided that shows exactly what gets hashed

src/hashing.rs
```diff
@@ -0,0 +1,494 @@
...
+            dst: PhantomData::<M>::default(),
+        };
+        result.update(M::domain_separation_tag(&label).as_bytes());
+        result
+    }
+
+    /// Adds the data to the digest function by first appending the length of the data in the byte array, and then
+    /// supplying the data itself.
+    pub fn update(&mut self, data: impl AsRef<[u8]>) {
+        self.inner.update(data.as_ref().len().to_le_bytes());
```
stringhandler, line 208: Perhaps change to either `u32` or `u64`, otherwise might be different on different targets depending on `usize`
```suggestion
        self.inner.update(data.as_ref().len().to_le_bytes());
```

src/ristretto/ristretto_keys.rs
```diff
@@ -240,6 +244,19 @@ impl RistrettoPublicKey {
         })
     }
 
+    /// A verifiable group generator using a domain separated hasher
+    pub fn group_generator(label: &str) -> Result<RistrettoPublicKey, HashingError> {
+        // This function requires 512 bytes of data, so let's be opinionated here and use blake2b
+        let hash = DomainSeparatedHasher::<Blake2b, RistrettoGroupGenerator>::new(label).finalize();
+        if hash.as_ref().len() < 64 {
```
stringhandler, line 251: Is this possible?

src/ristretto/ristretto_keys.rs
```diff
@@ -250,6 +267,36 @@ impl RistrettoPublicKey {
...
+    }
+
+    fn domain() -> &'static str {
+        "com.tari.kdf.ristretto"
+    }
+}
+
+pub struct RistrettoGroupGenerator;
+
+impl DomainSeparation for RistrettoGroupGenerator {
```
stringhandler, line 290: In some places we use Groups and some places PublicKey, should we not use one term for both?

stringhandler (commented):
Very nice. Happy to merge and create issues for the most. I think the `usize` issue should be fixed now though

src/ristretto/ristretto_keys.rs
CjS77: It's dependent on Blake2b obv. Sarang had the check in. Middle road is using assert_dbg

src/ristretto/ristretto_keys.rs
CjS77: Yeah, historical naming from the PoC.  Even in the MW context, don't we refer to `G`, `H` etc as generator points? Maybe `RistrettoGeneratorPoint`  is the best name

src/hashing.rs
CjS77: Good catch.
Should be `u64`

src/ristretto/ristretto_keys.rs
CjS77: On second thought, it's just a trivial amount of overhead that it doesn't hurt keeping it in, in case blake's impl changes, or someone wants to refactor in future.

stringhandler: Ok cool

AaronFeickert: Yeah, including the check is intended to make it clear what the output requirement is, and reduce the likelihood that future changes introduce a weakness.

src/hashing.rs
```diff
@@ -0,0 +1,495 @@
...
+impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
+    pub fn new<S>(label: S) -> Self
+    where S: AsRef<str> {
+        let inner = D::new();
+        let mut result = Self {
+            inner,
+            label: label.as_ref().to_string(),
+            dst: PhantomData::<M>::default(),
+        };
+        result.update(M::domain_separation_tag(&label).as_bytes());
```
sdbondi, line 201: @CjS77  @AaronFeickert made this comment in one of the discussions:

> Provided the method used for including the label and message in D is secure (simple concatenation, for example, is not sufficient), D_1 and D_2 behave as independent high-quality cryptographic hash functions, and generally retain the useful properties of D

I may be misunderstanding, but aren't we "simply concatenating (prepending)" the label to the message? Personalization in the Blake2b crate seems to be accomplished by passing a `persona` to `with_params`.

for e.g. 
```rust
VarBlake2b::with_params(
  //  key
  &[],
  // salt
  &[]
  // persona - must be 32 bytes
  b"com.tari.custom-32-byte-string_"
)
```

Or is this not required?

src/hashing.rs
```diff
@@ -0,0 +1,495 @@
...
+/// The domain separation information is retained with the hash, and can be queried using
+/// [`DomainSeparatedHash::domain_separation_tag`].
+///
+/// To preserve the guarantee that the hash is properly domain separated, you cannot create an instance of this struct
+/// directly. It is the result of using [`DomainSeparatedHasher`].
+///
+/// For details and examples, see [`DomainSeparatedHasher`].
+pub struct DomainSeparatedHash {
+    hash: Vec<u8>,
+    tag: String,
```
sdbondi, line 80: Both of these fields are heap-allocated. Since hashing is the bread and butter of crypto software, ideally these allocations should be avoided. `digest` provides the `Output<D> where D: Digest` type (alias to `generic-array` crate) that is stack-based and avoids the heap. 

As for the `tag`, perhaps it can be removed entirely as it probably won't ever be needed after hashing (if that info is really needed it could be preserved some other way e.g. a wrapper/new-type).

src/hashing.rs
```diff
@@ -0,0 +1,495 @@
...
+            label: label.as_ref().to_string(),
+            dst: PhantomData::<M>::default(),
+        };
+        result.update(M::domain_separation_tag(&label).as_bytes());
+        result
+    }
+
+    /// Adds the data to the digest function by first appending the length of the data in the byte array, and then
+    /// supplying the data itself.
+    pub fn update(&mut self, data: impl AsRef<[u8]>) {
```
sdbondi, line 207: I think this function should move to `impl digest::Update for DomainSeparatedHasher` trait (which gives `chain` for free)

src/hashing.rs
AaronFeickert: Safe label prepending is intended to be equivalent in functionality to the `Blake2b` `persona` parameter, but not every hash function's common APIs support this natively.

What I meant by "simple concatenation" being insufficient is that without proper handling of variable-length input, you can obtain collisions. Consider two examples that use concatenation
```
label_1 = "cats"
m_1 = "dogs"

label_2 = "cat"
m_2 = "sdogs"

D(label_1 || m_1) = D(label_2 || m_2) = D("catsdogs") // collision!
```

AaronFeickert: In the API that @CjS77 wrote, each label's length is prepended before the label itself, so there's no doubt where the label ends and subsequent inputs begin.

src/hashing.rs
CjS77: https://github.com/tari-project/tari-crypto/issues/113

src/hashing.rs
sdbondi: Ok thanks for clearing that up - makes sense

PR #111, opened 2022-06-24
feat!: change to nums constants

hansieodendaal:
Breaking change - any project based on a specific commitment in `tari_crypto` needs to be updated, as a commitment for the same value and blinding factor will now yield a different commitment. This was a necessary change as as described below.

- Changed NUMS constants used for the generators to make use of domain separated labels and not use iterative hashing of the Ristretto base point.
- Version update to `0.13.4` (this will align the version and tag as the previous version is `0.13.2` and the previous tag is `0.13.3`)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ categories = ["cryptography"]
 homepage = "https://tari.com"
 readme = "README.md"
 license = "BSD-3-Clause"
-version = "0.13.2"
+version = "0.13.4"
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -7,7 +7,7 @@ categories = ["cryptography"]
 homepage = "https://tari.com"
 readme = "README.md"
 license = "BSD-3-Clause"
-version = "0.13.2"
+version = "0.13.4"
```
stringhandler, line 10: Don't update the version, this is done in the tagging process

stringhandler, line 10:
```suggestion
version = "0.13.2"
```

stringhandler (dismissed):
Seems fine.

stringhandler (approved):

Cargo.toml
hansieodendaal: Ok cool

PR #116, opened 2022-07-08
fix(hashing): avoid heap allocations for domain-separated hashing API

### LINKED ISSUE #113: Avoid heap allocations in hashing API
As per #106,

Look into removing heap allocation for the hash output and string.

>Both of these fields are heap-allocated. Since hashing is the bread and butter of crypto software, ideally these allocations should be avoided. digest provides the Output<D> where D: Digest type (alias to generic-array crate) that is stack-based and avoids the heap.

>As for the tag, perhaps it can be removed entirely as it probably won't ever be needed after hashing (if that info is really needed it could be preserved some other way e.g. a wrapper/new-type).

sdbondi:
Hashing is an extremely common operation and therefore should avoid using the heap. 
This PR removes the allocation of the hash, the full tag string, and the label string from the `DomainSeparatedHasher`.  

~~BREAKING CHANGE: the version byte was previously converted to a decimal string representation before being applied to the digest. In order to maintain the same hashes, an allocation would necessarily`*` be required i.e. `let v_str = Self::version().to_string()`. Instead, the version byte is used directly in the challenge.~~

~~`*` barring some fancy use of `std::fmt::Formatter`~~

Fixes #113

New commit:[b591398](https://github.com/tari-project/tari-crypto/pull/116/commits/b591398c7e634b4e025a081614e51e8d65271201) Generates the ASCII byte representation for the version number without a heap allocation.

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -76,9 +76,9 @@ pub trait DomainSeparation {
     }
 }
 
-/// Converts a byte value to ASCII bytes that represent its value. This function returns a tuple containing the
-/// inclusive index of the most significant decimal value byte. For example,
-/// byte_to_decimal_ascii_bytes(0) returns (2, [0, 0, 48]).
+/// Converts a byte value to ASCII bytes that represent its value in big-endian order. This function returns a tuple
+/// containing the inclusive index of the most significant decimal value byte, and the 3 ASCII bytes (big-endian). For
+/// example, byte_to_decimal_ascii_bytes(0) returns (2, [0, 0, 48]).
 /// byte_to_decimal_ascii_bytes(42) returns (1, [0, 52, 50]).
 /// byte_to_decimal_ascii_bytes(255) returns (0, [50, 53, 53]).
 fn byte_to_decimal_ascii_bytes(mut byte: u8) -> (usize, [u8; 3]) {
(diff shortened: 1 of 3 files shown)


src/hashing.rs
```diff
@@ -129,81 +133,81 @@ impl AsRef<[u8]> for DomainSeparatedHash {
...
 ///
-/// fn card_id(card: &Card) -> DomainSeparatedHash {
+/// fn card_id(card: &Card) -> DomainSeparatedHash<Sha256> {
 ///     DomainSeparatedHasher::<Sha256, GenericHashDomain>::new("card_id")
 ///         .chain(card.name.as_bytes())
 ///         .chain(&[card.strength])
 ///         .finalize()
 /// }
 ///
+/// assert_eq!(GenericHashDomain::domain_separation_tag("card_id"), "com.tari.generic.v\u{1}.card_id");
```
CjS77, line 152: I'm sure we can do better here. Let's get the tag to be unchanged.

src/hashing.rs
```diff
@@ -287,31 +290,35 @@ impl DomainSeparation for MacDomain {
 /// ```
 ///
 /// ```
-/// use sha3::Sha3_256;
-/// use tari_crypto::hashing::Mac;
-/// use tari_utilities::hex::to_hex;
+/// # use sha3::Sha3_256;
+/// # use tari_crypto::hashing::{Mac, MacDomain, DomainSeparation};
+/// # use tari_utilities::hex::to_hex;
```
CjS77, line 295: I usually leave these in, especially the first example on a page, because it helps to know what you're bringing into the namespace when reading examples.

CjS77 (commented):
Untested ACK.

It would be nice to keep the tag 100% human-readoable though.

sdbondi:
@CjS77 yeah it's not ideal, could write up a byte to decimal conversion that does not allocate. Will do that

sdbondi:
@CjS77  I have updated the `add_domain_separation_tag` function to represent the version bytes as ASCII characters without any heap allocations. All hard-coded hashes in the tests have been reverted to `main`.

CjS77 (approved):
LGTM

Issue #110, opened 2022-06-24
Blake256 needs bounds checking on salt and persona bytes length


hansieodendaal:
`VarBlake2b::with_params` will throw an exception if either `salt` or `persona` bytes length exceed 16 bytes.

AaronFeickert:
@hansieodendaal: Looks like this issue can be closed via [PR 117](https://github.com/tari-project/tari-crypto/pull/117).

PR #117, opened 2022-07-11
fix: add checking for Blake256 parameters

DenisKolodin:
It adds checking for `Blake256` parameters: salt and persona should be less or equal 16 bytes length.

Motivation: the fixes for #110.

### DIFF
--- a/src/hash/blake2.rs
+++ b/src/hash/blake2.rs
@@ -5,7 +5,7 @@
 
 use blake2::{digest::VariableOutput, VarBlake2b};
 use digest::{
-    consts::U32,
+    consts::{U32, U64},
     generic_array::{typenum::Unsigned, GenericArray},
     FixedOutput,
     Reset,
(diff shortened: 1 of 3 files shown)


src/hash/blake2.rs
```diff
@@ -12,19 +12,25 @@ use digest::{
...
-    pub fn with_params(key: &[u8], salt: &[u8], persona: &[u8]) -> Self {
-        Self(VarBlake2b::with_params(
-            key,
-            salt,
-            persona,
-            <Self as FixedOutput>::OutputSize::USIZE,
-        ))
+    pub fn with_params(key: &[u8], salt: &[u8], persona: &[u8]) -> Result<Self, HashError> {
+        if salt.len() > 16 || persona.len() > 16 {
+            Err(HashError::WorngLength)
```
hansieodendaal, line 25:
```suggestion
            Err(HashError::WrongLength)
```

hansieodendaal (dismissed):
LGTM

hansieodendaal:
I prop0se we add another function to mae full use of the flexibility provided by `VarBlake2b::with_params`

``` rust
pub fn with_params_var_size(key: &[u8], salt: &[u8], persona: &[u8], output_size: usize) -> Result<Self, HashError>
```

src/hash/blake2.rs
```diff
@@ -5,26 +5,37 @@
...
-            salt,
-            persona,
-            <Self as FixedOutput>::OutputSize::USIZE,
-        ))
+    pub fn with_params(key: &[u8], salt: &[u8], persona: &[u8]) -> Result<Self, HashError> {
+        Self::with_params_var_size(key, salt, persona, <Self as FixedOutput>::OutputSize::USIZE)
+    }
+
+    /// Constructs a `Blake256` hashing context with an explicitly specified output size.
+    pub fn with_params_var_size(
```
hansieodendaal, line 28: I think we need a unit test for this, with all the edge cases.

stringhandler (approved):

PR #119, opened 2022-07-25
feat: implement Digest for DomainSeparatedHasher

stringhandler:
Implements `Digest` for DomainSeparatedHasher with a blank default label
Deleted `GenericHashDomain`

Motivation
---
This means you can use a DomainSeparatedHasher in any function that already uses Digest. Unfortunately, the label must be blank, so the `domain` should be different for struct.

I also added a macro `hashing_domain` for creating domains quickly (e.g. for tests)

```rust
hash_domain!(MyDemoHasher, "macro.test");
// Generates

 pub struct MyDemoHasher {}

        impl DomainSeparation for MyDemoHasher {
            fn version() -> u8 {
                1
            }

            fn domain() -> &'static str {
                "macro.test"
            }
        }   

   
``` 

After adding the `hash_domain` macro it is now easy enough to create a new domain when you need it, removing the need for `GenericHashDomain`. I found that devs were trying to build on top of it instead of creating new domains, so IMO it's more confusing keeping it in     

UPDATE: Removed `GenericHashDomain`
UPDATE: Added `create_hasher!` macro

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -254,6 +254,49 @@ impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
     }
 }
 
+/// Implements Digest so that it can be used for other crates
+impl<TInnerDigest: Digest, TDomain: DomainSeparation> Digest for DomainSeparatedHasher<TInnerDigest, TDomain> {
+    type OutputSize = TInnerDigest::OutputSize;
+
+    fn new() -> Self {
+        DomainSeparatedHasher::<TInnerDigest, TDomain>::new("")
+    }
+
+    fn update(&mut self, data: impl AsRef<[u8]>) {
+        self.update(data);
+    }
+
+    fn chain(self, data: impl AsRef<[u8]>) -> Self
+    where Self: Sized {
+        self.chain(data)
+    }
+
+    fn finalize(self) -> Output<Self> {
+        self.finalize().output
+    }
+
+    fn finalize_reset(&mut self) -> Output<Self> {
+        let value = self.inner.finalize_reset();
+        TDomain::add_domain_separation_tag(&mut self.inner, "");
+        value
+    }
+
+    fn reset(&mut self) {
+        self.inner.reset();
+        TDomain::add_domain_separation_tag(&mut self.inner, "");
+    }
+
+    fn output_size() -> usize {
+        TInnerDigest::output_size()
+    }
+
+    fn digest(data: &[u8]) -> Output<Self> {
+        let mut hasher = Self::new("");
+        hasher.update(data);
+        hasher.finalize().output
+    }
+}
+
 //-----------------------------------------    Generic Hash Domain  ----------------------------------------------------
 /// A domain separation marker for use in general use cases.
 pub struct GenericHashDomain;
@@ -427,6 +470,27 @@ pub trait DerivedKeyDomain: DomainSeparation {
     }
 }
 
+/// Creates a DomainSeparation struct for a given domain.
+#[macro_export]
+macro_rules! hash_domain {
+    ($name:ident, $domain:expr, $version: expr) => {
+        pub struct $name {}
+
+        impl DomainSeparation for $name {
+            fn version() -> u8 {
+                $version
+            }
+
+            fn domain() -> &'static str {
+                $domain
+            }
+        }
+    };
+    ($name:ident, $domain:expr) => {
+        hash_domain!($name, $domain, 1);
+    };
+}
+
 #[cfg(test)]
 mod test {
     use blake2::Blake2b;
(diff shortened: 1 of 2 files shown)


src/hashing.rs
```diff
@@ -254,6 +254,49 @@ impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
     }
 }
 
+/// Implements Digest so that it can be used for other crates
```
CjS77, line 257: The `digest` crate already offers a default impl for 

```
impl<D: FixedOutput + Default + Update + HashMarker> Digest for D 
```

Is this intended to be a stopgap so `digest` is available while we look at impl the other trait bounds?

stringhandler: I suppose I could implement `Update` and `Reset` individually instead, but I am not sure that covers all of them

stringhandler: I tried implementing them individually, but because we need to push a prefix when we reset, the `Reset` needs `Update`, and `FixedOutput` needs `Reset` and it ended up being a lot more code than implementing the whole of `Digest`.

hansieodendaal (approved):
Hi @stringhandler, LGTM, but please consider to merge [this PR](https://github.com/stringhandler/tari-crypto/pull/3) into your branch.

src/hashing.rs
```diff
@@ -468,10 +575,50 @@ mod test {
...
+            "5faa7d48b551362bbee8a02c43e6ab634ed47c58ecf7b353f9afedfe3d574608",
+        );
+
+        let mut hasher = DomainSeparatedHasher::<Blake256, MyDemoHasher>::new("");
+        hasher.update(&[0, 0, 0]);
+        let hash = hasher.finalize();
+        assert_eq!(
+            to_hex(hash.as_ref()),
+            "5faa7d48b551362bbee8a02c43e6ab634ed47c58ecf7b353f9afedfe3d574608"
+        );
```
hansieodendaal, line 592: I think it is important to show that this equality only holds if an empty label is supplied for the same hash domain, thus add a failing test for a hasher with a label.

Otherwise LGTM

hansieodendaal (approved):

src/hashing.rs
```diff
@@ -427,6 +461,48 @@ pub trait DerivedKeyDomain: DomainSeparation {
...
+        hasher
+    }};
+    ($digest: ty, $name:ident, $domain:expr, $label:expr) => {
+        create_hasher!($digest, $name, $domain, $label, 1)
+    };
+    ($digest: ty, $name:ident, $domain:expr) => {
+        create_hasher!($digest, $name, $domain, "", 1)
+    };
+    ($digest: ty, $domain:expr) => {
+        create_hasher!($digest, __TempHashDomain, $domain, "", 1)
```
sdbondi, line 502: What is `__TempHashDomain`?

src/hashing.rs
```diff
@@ -427,6 +461,48 @@ pub trait DerivedKeyDomain: DomainSeparation {
...
+    };
+}
+
+/// Creates a domain separated hasher.
+#[macro_export]
+macro_rules! create_hasher {
+    ($digest:ty, $name:ident, $domain:expr, $label:expr, $version: expr) => {{
+        use $crate::hash_domain;
+
+        hash_domain!($name, $domain, $version);
```
sdbondi, line 491: My only concern about `create_hasher!` is that users will duplicate the domain separation domain label.

```rust
fn hash_foo(foo:&Foo) -> Hash {
   let hasher = create_hasher!(Blake256, FooBarHasher, "com.foobar", "bar");
   hasher.chain(b"blah").finalize()
}

fn hash_bar(bar:&Bar) -> Hash {
   let hasher = create_hasher!(Blake256, FooBarHasher, "com.foobar", "bar");
   hasher.chain(b"blah").finalize()
}
```

Perhaps `create_hasher!` should take in a domain type?

```rust
define_hash_domain!(FooBarHashDomain, "com.foobar");

fn hash_foo(foo:&Foo) -> Hash {
   let hasher = create_hasher!(Blake256, FooBarHashDomain, "bar");
   hasher.chain(b"blah").finalize()
}

fn hash_bar(bar:&Bar) -> Hash {
   let hasher = create_hasher!(Blake256, FooBarHashDomain, "bar");
   hasher.chain(b"blah").finalize()
}
```

sdbondi (commented):
Implementing `Digest` and the `hash_domain!` macro is a great improvement - I have a concern about how `create_hasher!` should be used.

src/hashing.rs
stringhandler: The macro isn't really needed unless you are creating a new struct. (The DomainSeparation).

In your second example you can just create a method 

```
fn create_hasher<D:Digest, Dom:DomainSeparation>(label: &'static str) -> ...
```

stringhandler: Here's an example of `create_hasher!`, but I agree it's probably better to leave it out and create the method.

```rust 
     create_hasher!(Blake2b, "com.tari.cipher_seed_mac")
            .chain(plaintext.clone())
            .chain([CIPHER_SEED_VERSION])
            .chain(self.salt)
            .chain(passphrase.as_bytes())
            .finalize()
```

agubarev: same concern as @sdbondi, it's nice to add a suggestive comment not the use the long form macro `create_hasher!()`, maybe add another macro next to it, that would differ by name instead of only arguments

agubarev: users might not even notice that calling the long form in multiple places will define nested structs in different places where it's called

CjS77 (approved):

PR #120, opened 2022-07-26
feat: add digest convenience function to hashing api

hansieodendaal:
Added a digest convenience function to the hashing API


hansieodendaal:
Closed in favour of #123

PR #123, opened 2022-07-28
feat: synchronize the standard digest `new()` with `domain separated hasher::new()`

hansieodendaal:
With applying the new hashing API in the tari project some some suggestions are made to improve its use-ability:

- Renamed `DomainSeparatedHasher::<D, M>::new()` to   `DomainSeparatedHasher::<D, M>::new_with_label(label: &'static str)`
- Added `DomainSeparatedHasher<D, M>::new()` to to be functionally equivalent to `D::new()` due to use of `<DomainSeparatedHasher<D, M> as Digest>`
- Added `pub trait HashToBytes`
- Added a digest convenience function to the hashing API

_**Note:** The new error type `SliceError` must be `serde::{Deserialize, Serialize}` compatible and can therefor not be included in `HashingError`_

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -233,12 +233,7 @@ pub struct DomainSeparatedHasher<D, M> {
 impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
     /// Create a new instance of [`DomainSeparatedHasher`] without an additional label (to correspond to 'D::new()').
     pub fn new() -> Self {
-        let inner = D::new();
-        Self {
-            inner,
-            label: "",
-            _dst: PhantomData,
-        }
+        Self::new_with_label("")
     }
 
     /// Create a new instance of [`DomainSeparatedHasher`] for the given label.
@@ -281,7 +276,7 @@ impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
 }
 
 /// Convert a finalized hash into a fixed size buffer.
-pub trait HashToBytes<const I: usize>: AsRef<[u8]> {
+pub trait AsFixedBytes<const I: usize>: AsRef<[u8]> {
     /// A convenience function to convert a finalized hash into a fixed size buffer.
     fn as_fixed_bytes(&self) -> Result<[u8; I], SliceError> {
         let hash_vec = self.as_ref();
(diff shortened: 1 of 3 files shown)


src/hashing.rs
```diff
@@ -258,14 +272,37 @@ impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
...
+
+    /// A convenience function to update, then finalize the hasher and return the hash result.
+    pub fn digest(mut self, data: &[u8]) -> DomainSeparatedHash<D> {
+        self.update(data);
+        self.finalize()
+    }
 }
 
+/// Convert a finalized hash into a fixed size buffer.
+pub trait HashToBytes<const I: usize>: AsRef<[u8]> {
```
stringhandler, line 284:
```suggestion
pub trait AsFixedBytes<const I: usize>: AsRef<[u8]> {
```

src/hashing.rs
```diff
@@ -541,18 +583,18 @@ mod test {
             hasher!(Blake256, MyDemoHasher, "com.macro.test");
 
             util::hash_from_digest(
-                MyDemoHasher::new(""),
+                MyDemoHasher::new(),
                 &[0, 0, 0],
-                "5faa7d48b551362bbee8a02c43e6ab634ed47c58ecf7b353f9afedfe3d574608",
+                "a980ec75b9b1a6b2bc50884e01b319c9184bef53ef7906ee7c1fb9afd96a377b",
```
stringhandler, line 588: This should not have changed

src/hashing.rs
```diff
@@ -227,8 +231,18 @@ pub struct DomainSeparatedHasher<D, M> {
 }
 
 impl<D: Digest, M: DomainSeparation> DomainSeparatedHasher<D, M> {
+    /// Create a new instance of [`DomainSeparatedHasher`] without an additional label (to correspond to 'D::new()').
+    pub fn new() -> Self {
```
stringhandler, line 235: This is not correct. It does not call `M::add_domain_separation_tag`
```suggestion
    pub fn new() -> Self {
       Self::new_with_label("")
```

stringhandler (changes requested):

src/hashing.rs
hansieodendaal: O yes... good catch

stringhandler (approved):

PR #125, opened 2022-08-04
feat: add domain separated hashing to wasm

hansieodendaal:
Added domain separated hashing to the wasm bindings.

**Note:** 
Complex coding using mutable lazy statics and unsafe blocks with memory transmutation was required to get this to work. Maybe there exists a better solution than this without changing the `DomainSeparation` trait? One thing to consider could be [this comment](https://github.com/tari-project/tari-crypto/pull/106#issuecomment-1169335255) in #106.

_Thanks to @DenisKolodin for assisting with `unsafe transmute` code._

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,6 +31,7 @@ sha3 = "0.9.0"
 thiserror = "1.0.20"
 wasm-bindgen = { version = "^0.2", features = ["serde-serialize"], optional = true }
 zeroize = "1.4.0"
+mut_static = "5.0.0"
 
 [dev-dependencies]
 bincode = "1.1.4"
--- a/src/wasm/hashing.rs
+++ b/src/wasm/hashing.rs
@@ -1,0 +1,190 @@
+// Copyright 2020. The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! Hashing API
+
+use mut_static::MutStatic;
+use serde::{Deserialize, Serialize};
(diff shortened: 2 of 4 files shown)


DenisKolodin:
In fairness to say that `unsafe` in not a desired approach, but a workaround. To avoid using that unsafe code we one of the following ideas:

1. Add a lifetime to the `DomainSeparation` to let it to exist in the narrow scope:

```rust
trait DomainSeparation<'a> {
    fn version(&'a self) -> u8;
    fn domain(&'a self) -> &'a str;
}

struct Mac;

impl DomainSeparation<'static> for Mac {
    fn version(&'static self) -> u8 {
        8
    }
    fn domain(&'static self) -> &'static str {
        "domain.com"
    }
}

fn main() {
    format!("{}", Mac.domain());
}
```

2. Use a struct offered here: https://github.com/tari-project/tari-crypto/pull/106#issuecomment-1169335255

3. Declare domains in the Rust code and access them by numbers/constants.

stringhandler (commented):
I'm not a fan of this approach. The DomainSeparation trait is used to prevent mistakes in Rust through compilation errors, so will not lend itself to the dynamic nature of JS and WASM. 

I suspect we will need an overload or trait at the DomainSeparatedHasher level, or even to construct the hashers manually.

Where is this actually getting used? Perhaps an example will help clarify the need.

hansieodendaal:
Ok, parking it for now

PR #126, opened 2022-08-04
fix: zeroize temp fields during serializing

stringhandler:


DenisKolodin (approved):

CjS77:
Needs a cargo fmt. Otherwise +1 here too

sdbondi (approved):

PR #130, opened 2022-09-23
fix: update commitment signature

AaronFeickert:
The commitment signature is special sound as a zero-knowledge proof of knowledge of the opening of a commitment, and can be bound to an arbitrary message for use as a signature construction.

However, in practice there are two input commitments used by the prover and verifier: a standard Pedersen commitment and an additional commitment to zero (either a sender offset or script public key). These commitments are summed and, while they are included separately in the Fiat-Shamir transcript, successful verification does not imply prover knowledge of both commitment openings separately.

While it is possible to update the commitment signature construction using a standard Chaum-Pedersen construction, this would increase the signature size. Another approach is to use a power-of-challenge design that retains the same signature size. This PR uses the latter approach, and updates tests and documentation accordingly. The commitment to zero is handled as a public key, but it remains the caller's responsibility to ensure it uses the correct generator. As before, the Fiat-Shamir transcript does _not_ bind to generators, assuming they are public and globally fixed.

It is worth noting that a future use of the commitment signature may not need to prove knowledge of the opening of a commitment to zero; in this case, it suffices to use zero values for the public key and corresponding witness.

### DIFF
--- a/src/signatures/commitment_signature.rs
+++ b/src/signatures/commitment_signature.rs
@@ -24,26 +24,21 @@ pub enum CommitmentSignatureError {
     InvalidChallenge,
 }
 
-/// # Commitment Signatures
+/// # Commitment signatures
 ///
-/// Find out more about Commitment signatures [here](https://eprint.iacr.org/2020/061.pdf) and
-/// [here](https://documents.uow.edu.au/~wsusilo/ZCMS_IJNS08.pdf).
+/// A commitment signature is a zero-knowledge proof of knowledge (bound to a message as a signature) of the
+/// opening of a commitment and, optionally, another commitment to zero (interpreted as a public key).
 ///
-/// In short, a Commitment Signature is made up of the tuple _(R, u, v)_, where _R_ is a random Pedersen commitment (of
-/// two secret nonces) and _u_ and _v_ are the two publicly known private signature keys. It demonstrates ownership of
-/// a specific commitment.
+/// This is used elsewhere in Tari protocols to authorize transactions.
 ///
-/// The Commitment Signature signes a challenge with the value commitment's value and blinding factor. The two nonces
-/// should be completely random and never reused - that responsibility lies with the calling function.
-///   C = a*H + x*G          ... (Pedersen commitment to the value 'a' using blinding factor 'x')
-///   R = k_2*H + k_1*G      ... (a public (Pedersen) commitment nonce created with the two random nonces)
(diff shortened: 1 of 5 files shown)


src/signatures/commitment_signature.rs
```diff
@@ -24,26 +24,21 @@ pub enum CommitmentSignatureError {
...
+/// uses a power-of-challenge design for size efficiency.
 ///
-/// Verification of the Commitment Signature (R, u, v) entails the following:
-///   S = v*H + u*G          ... (Pedersen commitment of the publicly known private signature keys)
-///   S =? R + e.C           ... (final verification)
+/// For a commitment `C = a*H + x*G` and commitment to zero `D = y*G`, the proving system protocol is:
+/// - The prover selects random nonces `k_1` and `k_2`, and builds the ephemeral commitment `R = k_2*H + k_1*G`.
+/// - The verifier receives `C`, `D`, and `R`, and selects a random challenge `e`.
+/// - The prover computes `u = k_1 + e*x + e*e*y` and `v = k_2 + e*a`.
+/// - The verifier receives `u` and `v`, and accepts the proof if and only if `v*H + u*G = R + e*C + e*e*D`.
```
hansieodendaal, line 41: So for me to understand this properly, with reference to [RFC-0201_TariScript](https://rfc.tari.com/RFC-0201_TariScript.html):
- For the `metadata_signature` in the transaction output, (5) [_with commitment openings_] belongs to the receiver, and  (6)/(11) [_with sender offset private key_] belongs to the sender, so I am sure how this will work?
- For the `script_signature` in the transaction input the sender knows all secrets (commitment openings and script private key), so I can see how the single party can construct the signature.
- In both cases the challenge cannot be random as it either has to bind all metadata, is this ok?

AaronFeickert: A similar multiparty computation works for the power-of-challenge construction as for the current implementation. The only difference is how the sender computes its response value (it uses a different power of the challenge).

Sorry about any confusion on the challenge computation. In the interactive proving system, the challenge is sampled randomly by the verifier. In the non-interactive version (which is used here), the challenge is computed using Fiat-Shamir.

AaronFeickert:
While working out the security proof for this construction, it became evident that it can only show that both input values (the commitment and claimed commitment to zero) have known representations with respect to both generators, which does not meet the desired requirements.

Fortunately, an alternate approach will do the trick. Until then, closing this PR so it isn't merged.

PR #131, opened 2022-09-29
feat: add new commitment signature to use complete representation proof

AaronFeickert:
Commitment signatures are currently used (via the `CommitmentSignature` type) in Tari protocols as part of transaction authorization, as described in [RFC-0201](https://rfc.tari.com/RFC-0201_TariScript.html). The cryptographic design of commitment signatures is such that they are a representation proof of a Pedersen commitment that is bound to arbitrary message data (via a Fiat-Shamir challenge) to produce a signature.

In both cases where commitment signatures are used, two elements of input data are summed prior to generating or verifying the signature. The cases are:
- an output commitment and a sender offset public key
- an input commitment and a script public key
In both cases, the individual elements are included separately in the challenge.

This approach is not a proof of knowledge of the openings of the individual elements. Instead, it is a proof of knowledge of the commitment value, and of the sum of the commitment mask and discrete logarithm of the public key. It is unclear if this can lead to practical transaction malleability or other insecure outcomes.

This PR adds a new `CommitmentAndPublicKeySignature` signature type that mitigates the issue by proving knowledge of the commitment value, commitment mask, and public key discrete logarithm. As before, it is the responsibility of the caller to ensure that the Fiat-Shamir challenge is generated correctly.

The existing `CommitmentSignature` is left unchanged, but should not be used for the aforementioned use cases.

### DIFF
--- a/src/signatures/commitment_and_public_key_signature.rs
+++ b/src/signatures/commitment_and_public_key_signature.rs
@@ -1,0 +1,349 @@
+// Copyright 2021. The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+use std::{
+    cmp::Ordering,
+    hash::{Hash, Hasher},
+    ops::{Add, Mul},
+};
+
+use rand::thread_rng;
+use serde::{Deserialize, Serialize};
+use tari_utilities::ByteArray;
+use thiserror::Error;
+
+use crate::{
+    commitment::{HomomorphicCommitment, HomomorphicCommitmentFactory},
+    keys::{PublicKey, SecretKey},
+};
+
+/// An error when creating a commitment signature
+#[derive(Clone, Debug, Error, PartialEq, Eq, Deserialize, Serialize)]
+#[allow(missing_docs)]
+pub enum CommitmentAndPublicKeySignatureError {
+    #[error("An invalid challenge was provided")]
+    InvalidChallenge,
+}
+
+/// # Commitment and public key (CAPK) signatures
+/// 
+/// Given a commitment `commitment = a*H + x*G` and group element `pubkey = y*G`, a CAPK signature is based on
+/// a representation proof of both openings: `(a, x)` and `y`. It additionally binds to arbitrary message data `m`
+/// via the challenge to produce a signature construction.
+/// 
+/// It is used in Tari protocols as part of transaction authorization.
+/// 
+/// The construction works as follows:
+/// - Sample scalar nonces `r_a, r_x, r_y` uniformly at random.
+/// - Compute ephemeral values `ephemeral_commitment = r_a*H + r_x*G` and `ephemeral_pubkey = r_y*G`.
+/// - Use strong Fiat-Shamir to produce a challenge `e`. If `e == 0` (this is unlikely), abort and start over.
+/// - Compute the responses `u_a = r_a + e*a` and `u_x = r_x + e*x` and `u_y = r_y + e*y`.
+/// 
+/// The signature is the tuple `(ephemeral_commitment, ephemeral_pubkey, u_a, u_x, u_y)`.
+/// 
+/// To verify:
+/// - The verifier computes the challenge `e` and rejects the signature if `e == 0` (this is unlikely).
+/// - Verification succeeds if and only if the following equations hold:
+///     `u_a*H + u*x*G == ephemeral_commitment + e*commitment`
+///     `u_y*G == ephemeral_pubkey + e*pubkey`
+/// 
+/// We note that it is possible to make verification slightly more efficient. To do so, the verifier selects a nonzero
+/// scalar weight `w` uniformly at random (not through Fiat-Shamir!) and accepts the signature if and only if the
+/// following equation holds:
+///     `u_a*H + (u_x + w*u_y)*G - ephemeral_commitment - w*ephemeral_pubkey - e*commitment - (w*e)*pubkey == 0`
+/// The use of efficient multiscalar multiplication algorithms may also be useful for efficiency.
+/// The use of precomputation tables for `G` and `H` may also be useful for efficiency.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct CommitmentAndPublicKeySignature<P, K> {
+    ephemeral_commitment: HomomorphicCommitment<P>,
+    ephemeral_pubkey: P,
+    u_a: K,
+    u_x: K,
+    u_y: K,
+}
+
+impl<P, K> CommitmentAndPublicKeySignature<P, K>
+where
+    P: PublicKey<K = K>,
+    K: SecretKey,
+{
+    /// Creates a new [CommitmentSignature]
+    pub fn new(ephemeral_commitment: HomomorphicCommitment<P>, ephemeral_pubkey: P, u_a: K, u_x: K, u_y: K) -> Self {
+        CommitmentAndPublicKeySignature { ephemeral_commitment, ephemeral_pubkey, u_a, u_x, u_y }
+    }
+
+    /// Complete a signature using the given challenge. The challenge is provided by the caller to support the
+    /// multiparty use case. It is _very important_ that it be computed using strong Fiat-Shamir! Further, the
+    /// values `r_a, r_x, r_y` are nonces, must be sampled uniformly at random, and must never be reused.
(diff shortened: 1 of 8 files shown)


src/signatures/commitment_signature.rs
```diff
@@ -24,32 +24,40 @@ pub enum CommitmentSignatureError {
...
+/// - The verifier computes the challenge `e` and rejects the signature if `e == 0` (this is unlikely).
+/// - Verification succeeds if and only if the following equations hold:
+///     `u_a*H + u*x*G == ephemeral_commitment + e*commitment`
+///     `u_y*G == ephemeral_pubkey + e*pubkey`
+/// 
+/// We note that it is possible to make verification slightly more efficient. To do so, the verifier selects a nonzero
+/// scalar weight `w` uniformly at random (not through Fiat-Shamir!) and accepts the signature if and only if the
+/// following equation holds:
+///     `u_a*H + (u_x + w*u_y)*G - ephemeral_commitment - w*ephemeral_pubkey - e*commitment - (w*e)*pubkey == 0`
+/// The use of efficient multiscalar multiplication algorithms may also be useful for efficiency.
```
hansieodendaal, line 53: This seems like a good suggestion.

src/signatures/commitment_signature.rs
```diff
@@ -85,88 +89,131 @@ where
...
-        // R + e.C
-        let rhs = &self.public_nonce + &(public_commitment * challenge);
-        // Implementors should make this a constant time comparison
-        lhs == rhs
+        // The challenge cannot be zero
+        if challenge.to_owned() == K::default() {
+            return false
+        }
+
+        // TODO: Use efficient multiscalar multiplication with a random verifier weight between the equations
```
hansieodendaal, line 157: Why is this `// TODO`? Is it not as straight forward as presented in the comment^^?

hansieodendaal (dismissed):
LGTM

I like the pragmatic approach shown in the tests and that the _commitment signature includes the commitment and additional public key and is based on a representation proof of both openings_. On the rare occasions that we do not have an additional public key default values can be used.

**Q / Nit**: That being said, maybe we can consider changing the interface slightly?
`AggregatedCommitmentSignature` could represent the case where we have a commitment and additional public key.
`CommitmentSignature` could represent the case where we do not have an additional public key, using exactly the same code but defaulting the public key inputs behind the scene.

src/signatures/commitment_signature.rs
AaronFeickert: It wasn't clear to me if it was straightforward or possible to use multiscalar multiplication evaluation given the generic key types, nor if it's even an efficiency win in the non-batched case.

AaronFeickert:
> LGTM
> 
> I like the pragmatic approach shown in the tests and that the _commitment signature includes the commitment and additional public key and is based on a representation proof of both openings_. On the rare occasions that we do not have an additional public key default values can be used.
> 
> **Q / Nit**: That being said, maybe we can consider changing the interface slightly? `AggregatedCommitmentSignature` could represent the case where we have a commitment and additional public key. `CommitmentSignature` could represent the case where we do not have an additional public key, using exactly the same code but defaulting the public key inputs behind the scene.

Thanks for the review!

I'm not sure what would be gained by separating into different signature types in this case. Responsibility for computing the challenge correctly via strong Fiat-Shamir is offloaded to the caller (an approach I don't particularly like), so the caller would still need to handle the difference between signature types itself, and bind to the appropriate statement and transcript values.

src/signatures/commitment_signature.rs
AaronFeickert: I'll be implementing it in a subsequent commit! I want to do a few efficiency tests first.

stringhandler (commented):
Agree with the concept, but I don't agree with the implementation.

The sign method here is very tightly coupled with the implementation in Tari that is signing using a commitment and a public key, but the ComSig in this library should be for general purpose signing with a commitment. I would suggest leaving this implementation as is, and creating a new struct (e.g. CommitmentAndPublicKeySignature) that represents the use case more clearly

AaronFeickert:
> Agree with the concept, but I don't agree with the implementation.
> 
> The sign method here is very tightly coupled with the implementation in Tari that is signing using a commitment and a public key, but the ComSig in this library should be for general purpose signing with a commitment. I would suggest leaving this implementation as is, and creating a new struct (e.g. CommitmentAndPublicKeySignature) that represents the use case more clearly

Is there a point in the protocol where it's known that a "standard" `CommitmentSignature` would be needed?

That being said, if keeping the intent of the use case clear is a design priority, we certainly could build a `CommitmentAndPublicKeySignature`. Heck, if someone were bored, they could generalize it to arbitrary representations of arbitrary sets of group elements (I've seen at least one design specification elsewhere that takes an insanely general approach), but at some point it seems unnecessary.

hansieodendaal (approved):
LGTM -nice and clean

AaronFeickert:
Can someone approve CI workflows for this PR? I don't have local Wasm testing working yet, so it's possible the last commit might still need work.

src/signatures/commitment_signature.rs
AaronFeickert: This is (partially) done in a more recent commit. Verification now merges the two equations into a single weighted equation, but still calls into the commitment factory to take advantage of future precomputation there. It would be most ideal to use precomputation with curve library multiscalar multiplication, but that can be left to future work if desired.

stringhandler:
This is more in line with what I had in mind.

src/signatures/commitment_and_public_key_signature.rs
```diff
@@ -0,0 +1,349 @@
...
+        // u_a*H + (u_x + w*u_y)*G == ephemeral_commitment + w*ephemeral_pubkey + e*commitment + (w*e)*pubkey
+        let verifier_lhs = factory.commit(&(&self.u_x + &(&w * &self.u_y)), &self.u_a).as_public_key().to_owned();
+        let verifier_rhs_unweighted = self.ephemeral_commitment.as_public_key() + (commitment * challenge).as_public_key();
+        let verifier_rhs_weighted = &self.ephemeral_pubkey * &w + pubkey * &(&w * challenge);
+
+        verifier_lhs == verifier_rhs_unweighted + verifier_rhs_weighted
+    }
+
+    /// Get the signature tuple `(ephemeral_commitment, ephemeral_pubkey, u_a, u_x, u_y)`
+    #[inline]
```
stringhandler, line 179: These inlines are not necessary, and IMO should only be added if they are proven to add performance

src/signatures/commitment_and_public_key_signature.rs
```diff
@@ -0,0 +1,349 @@
...
+        // The challenge cannot be zero
+        if *challenge == K::default() {
+            return false
+        }
+
+        // Use a single weighted equation for verification to avoid unnecessary group operations
+        // For now, we use naive multiscalar multiplication, but offload the commitment computation
+        // This allows for the use of precomputation within the commitment itself, which is more efficient
+        // TODO: use curve library (Straus/Pippenger) multiscalar multiplication with partial precomputation
+        let mut rng = thread_rng();
```
stringhandler, line 167: The RNG should be passed into the method as a parameter. Different platforms (e.g. WASM) may not have this RNG, so rather, add it as a parameter with the trait bound CryptoRng (I think this is the name)

stringhandler (commented):
Happy in general, so comments. Looks like there are also some matches that can be converted to ifs

src/signatures/commitment_and_public_key_signature.rs
AaronFeickert: Removed for both signature types.

AaronFeickert:
> Happy in general, so comments. Looks like there are also some matches that can be converted to ifs

Which ones specifically are you thinking of?

AaronFeickert:
For some reason, DeepSource only complained about some uses of `match` within the new code from this PR. It did not catch all such uses, nor did it flag similar uses from the original `CommitmentSignature` code. This means the code is currently inconsistent in this manner.

Recommend that either the linting commits be reverted and DeepSource's complaints be ignored, or the rest of the relevant `match` uses be updated for consistency.

hansieodendaal (dismissed):
utACK

stringhandler (approved):

PR #132, opened 2022-10-03
feat: port dalek

SWvheerden:
Move away for `curve25519-dalek-ng` and rather use `curve25519-dalek` as this is the more updated repo.
Updated the version as this is a breaking change with respect to dependencies.
Matched the version of `rand` to `rand_core` so that the traits can work.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,26 +11,26 @@ version = "0.15.5"
 edition = "2018"
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


CjS77:
Looks like benchmarks need updating too

CjS77:
..and the wasm :)

stringhandler (approved):

PR #134, opened 2022-10-04
fix: include wasm features only if required

sdbondi:
Only includes `rand/wasm-bindgen` when the `wasm` feature is active.
It's not clear why stdweb was added. Since some wasm targets include rust stdlib, or are no_std. If required, it should probably be behind another feature flag. EDIT: https://rustsec.org/advisories/RUSTSEC-2020-0056.html


SWvheerden (approved):
LGTM

CjS77 (approved):
LGTM

Issue #135, opened 2022-10-04
Expose and use precomputation tables for commitment generators


AaronFeickert:
Precomputation tables for fixed generators, like those used in commitments and public keys, can provide significant relative performance benefits in scalar-group multiplication operations. The "default" group generator `G` already has a table hardcoded in the curve library, and the API allows you to generate tables for any group element you wish. Aside from using tables for function-based operations like commitments, it may additionally be useful to expose references to these tables for higher-level protocol operations (like commitment signatures) that handle their own scalar-group multiplications.

An unfortunate downside of the `dalek` API is that it handles scalar-group multiplication precomputation differently than for multiscalar multiplication operations, and the latter is quite inflexible. There are a few options for how to handle these limitations, some of which might necessitate a custom fork of the curve library to expose some of the guts of the implementation.

For reference, local benchmarks suggest a 60% performance gain for generating a Pedersen commitment with the use of precomputation tables for both generators, relative to the use of constant-time multiscalar multiplication with no tables.

AaronFeickert:
One option for this could be to generate precomputation tables dynamically when a `PedersenCommitmentFactory` is created. However, this breaks the "it's cheap to create factories" guideline implied in the code comments.

Another option is to generate precomputation tables statically along with NUMS points. Most of them won't be used, however.

A middle ground could be statically creating only the standard `H` point's table with the NUMS points. Then creation of a `PedersenCommitmentFactory` could store references to it and the standard `G` point's table in the default case, or generate new tables in the non-default case. This would make the default case efficient even if the factory is created multiple times.

hansieodendaal:
https://github.com/AaronFeickert/tari-crypto/commit/5d0aa493ba2ddbdcda631cebd5b8eea20b6b76bf

PR #136, opened 2022-10-11
feat: use precomputation for default commitments

AaronFeickert:
This work adds precomputation support for Pedersen commitments that use a provided generator. Basepoint tables are created via `lazy_static` alongside the generators themselves. When computing a commitment using the default generators, the corresponding tables are used. Commitments using any other generators are unchanged.

Closes [issue #135](https://github.com/tari-project/tari-crypto/issues/135).

### DIFF
--- a/src/ristretto/constants.rs
+++ b/src/ristretto/constants.rs
@@ -1,7 +1,7 @@
 // Copyright 2019. The Tari Project
 // SPDX-License-Identifier: BSD-3-Clause
 
-//! Constant [NUMS](https://tools.ietf.org/id/draft-black-numscurves-02.html) points for the Ristretto curve. There are 10 provided, but this library currently only
+//! Constant [NUMS](https://www.ietf.org/archive/id/draft-black-numscurves-02.txt) points for the Ristretto curve. There are 10 provided, but this library currently only
 //! uses the first
 
 use curve25519_dalek::ristretto::{CompressedRistretto, RistrettoPoint, RistrettoBasepointTable};
--- a/src/ristretto/constants.rs
+++ b/src/ristretto/constants.rs
@@ -65,14 +65,8 @@ lazy_static! {
         arr
     };
 
-    /// Precomputation tables for the points
-    pub static ref RISTRETTO_NUMS_TABLES: Vec<RistrettoBasepointTable> = {
-        let mut arr = Vec::<RistrettoBasepointTable>::with_capacity(NUMBER_NUMS_POINTS);
-        for i in 0..NUMBER_NUMS_POINTS {
-            arr.push(RistrettoBasepointTable::create(&RISTRETTO_NUMS_POINTS[i]));
-        }
-        arr
-    };
+    /// Precomputation table for the first point, which is used as the default commitment generator
+    pub static ref RISTRETTO_NUMS_TABLE_0: RistrettoBasepointTable = RistrettoBasepointTable::create(&RISTRETTO_NUMS_POINTS[0]);
 }
 
 #[cfg(test)]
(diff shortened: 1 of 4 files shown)


AaronFeickert:
One downside to this implementation is that it generates tables for all provided points, but only checks for the default generators when computing Pedersen commitments. Checking for any combination of the standard basepoint (`G`) and one of the provided generators seems like unnecessary computation that won't target any practical use case.

Should we only generate a table for the default `H` point?

stringhandler (dismissed):

src/ristretto/constants.rs
```diff
@@ -62,17 +62,26 @@ lazy_static! {
         }
         arr
     };
+
+    /// Precomputation tables for the points
+    pub static ref RISTRETTO_NUMS_TABLES: Vec<RistrettoBasepointTable> = {
+        let mut arr = Vec::<RistrettoBasepointTable>::new();
```
sdbondi, line 68: A [lazy static](https://docs.rs/lazy_static/latest/lazy_static/) is perhaps better here. This code will always be invoked whenever the tari_crypto crate is used. Since there is computation and a heap allocation, I think doing this as needed/lazily makes sense.

```suggestion
        let mut arr = Vec::<RistrettoBasepointTable>::with_capacity(10);
```

sdbondi (commented):
LGTM - one minor comment

sdbondi (dismissed):
ACK

Thanks for adding a bench :)

src/ristretto/constants.rs
```diff
@@ -4,13 +4,15 @@
 //! Constant [NUMS](https://tools.ietf.org/id/draft-black-numscurves-02.html) points for the Ristretto curve. There are 10 provided, but this library currently only
```
jorgeantonio21, line 4: JFY, this link is broken. Can we update the reference please ?

src/ristretto/constants.rs
```diff
@@ -55,24 +57,27 @@ pub const RISTRETTO_NUMS_POINTS_COMPRESSED: [CompressedRistretto; 10] = [
...
+    pub static ref RISTRETTO_NUMS_POINTS: [RistrettoPoint; NUMBER_NUMS_POINTS] = {
+        let mut arr = [RistrettoPoint::default(); NUMBER_NUMS_POINTS];
+        for i in 0..NUMBER_NUMS_POINTS {
             arr[i] = RISTRETTO_NUMS_POINTS_COMPRESSED[i].decompress().unwrap();
         }
         arr
     };
+
+    /// Precomputation table for the first point, which is used as the default commitment generator
+    pub static ref RISTRETTO_NUMS_TABLE_0: RistrettoBasepointTable = RistrettoBasepointTable::create(&RISTRETTO_NUMS_POINTS[0]);
```
jorgeantonio21, line 69: If I understand correctly, we only need the first entry of `RISTRETTO_NUMS_POINTS` to defined `RISTRETTO_NUM_TABLE_0`, why do we need then an array `RISTRETTO_NUMS_POINTS_COMPRESSED` of len `10` ?

src/ristretto/constants.rs
AaronFeickert: Good catch! I had not tested the link to ensure it was still active.

src/ristretto/constants.rs
AaronFeickert: The library already contained this set of points for implementers to use. The default Pedersen factory uses the standard base point `G` (from the upstream curve library) and the first point in the set as its generators. The [extended Pedersen factories](https://github.com/tari-project/tari-crypto/blob/main/src/ristretto/pedersen/extended_commitment_factory.rs) use additional points from the set.

An earlier version of this PR generated precomputation tables for all the points in the set, but it was decided this was wasted computation at this point, since only the default Pedersen factory is used in the codebase (and the extended Pedersen factories use multiscalar multiplication for commitment evaluation, which would require different precomputation table structures and architecture).

AaronFeickert: @CjS77 had mentioned elsewhere that a future strategy could be to keep some expensive curve point generation operations (like hash-based generation and precomputation) within the commitment factories, and then ensure that these factories are generated only once within the implementing codebase. This is already how the much larger set of range proof generators is handled. However, default Pedersen factory copies are currently generated on the fly more often, presumably because it was cheap to do so.

src/ristretto/constants.rs
AaronFeickert: Fixed.

stringhandler (approved):

PR #137, opened 2022-10-21
feat: add `Zeroize` support to key types, and create new shared secret type

AaronFeickert:
Adds support for `Zeroize` to both the `RistrettoSecretKey` and `RistrettoPublicKey` types, and creates a new shared secret type. Adds a test for both key types.

This work adds derived `Zeroize` support to `RistrettoSecretKey` to support the use case where we want to clear a secret key in scope.

It also adds custom `Zeroize` support to `RistrettoPublicKey`. In this case, we zeroize both the underlying `RistrettoPoint` and the `CompressedRistrettoPoint` contained in a `OnceCell`. This is useful in the case where a `RistrettoPublicKey` represents secret data, like in the case of a Diffie-Hellman shared secret. Note that we do not zeroize on drop, so this use case must be handled manually, ideally through the use of a `Zeroizing` wrapper.

Finally, it adds a new generic `DiffieHellmanSharedSecret<P>` type for `P: Zeroize`. This type zeroizes on drop. It does not provide direct access to the underlying public key, but only gives an `as_bytes` array view. This removes the need for `zeroize` support by implementations, and provides safer use of shared secrets.

### DIFF
--- a/src/dhke.rs
+++ b/src/dhke.rs
@@ -1,0 +1,48 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! The robotic innards of a Diffie-Hellman key exchange (DHKE) producing a shared secret.
+//! Even though the result of a DHKE is the same type as a public key, it is typically treated as a secret value.
+//! To make this work more safely, we ensure that a DHKE result is cleared after use (but beware of subsequent copies or moves).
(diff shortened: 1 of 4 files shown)


CjS77 (dismissed):
LGTM

hansieodendaal:
utACK

stringhandler (approved):
utACK

PR #140, opened 2022-10-26
feat: deterministic pseudorandom operations

AaronFeickert:
This is a proof of concept for a deterministic shuffler and bounded number generator for use in consensus-critical applications. It provides a generic API that lets you choose your favorite CSPRNG.

Some protocol applications are likely to require the use of consensus-based pseudorandomness. For example, the network may need to choose or shuffle among validator nodes using chain state data as a common seed. This means that all such nodes must take this seed and use it to arrive at the same results. While `rand` has generic functionality for these kinds of operations using any random number generator, there is no guarantee that its algorithms for doing so will be the same across versions, which could lead to dependency issues.

This work provides one design that has no `rand` dependency. It offers a generic `DeterministicRandomizer` type that can be instantiated with any seedable CSPRNG. The tests use a `ChaCha12`-based generator that happens to be the standard one used in `rand` currently, but which we import explicitly from `rand_chacha` instead.

When a bounded `u32` is requested, it takes a `u64` value from the PRNG and reduces it. When a bounded `u64` is requested, it takes two `u64` values from the PRNG, uses bit shifting to produce a `u128` value, and reduces it. This approach mitigates bias.

Comments welcome.

### DIFF
--- a/src/deterministic_randomizer.rs
+++ b/src/deterministic_randomizer.rs
@@ -97,7 +97,7 @@ where
         self.seed = seed;
     }
 
-    /// Reset the randomizer to its original seed
+    /// Reset the randomizer using the seed last provided by either the constructor or reseeding
     pub fn reset(&mut self) {
         self.reseed(self.seed.clone());
     }
--- a/src/deterministic_randomizer.rs
+++ b/src/deterministic_randomizer.rs
@@ -1,0 +1,188 @@
+// Copyright 2022 The Tari Project
+// SPDX-License-Identifier: BSD-3-Clause
+
+//! A deterministic randomizer with utility functions for operating on numbers and arrays in a reproducible and
+//! platform-indepdent way.
+
+use std::convert::TryFrom;
+
+use rand_core::{CryptoRng, RngCore, SeedableRng};
+use zeroize::Zeroize;
+
+/// Error representing a failed shuffle
+#[derive(Debug)]
+pub struct RandomizerError;
+
+/// A `DeterministicRandomizer` contains utility functions for working with pseudorandom number generators (PRNGs) where
+/// reproducibility is important. While the `rand` crate has functionality for things like uniform sampling and
+/// shuffling that can be used with PRNGs, there is no guarantee that these functions will always be implemented in the
+/// same way across versions, which could lead to incompatibility or dependency troubles. This library has no specific
+/// `rand` dependency, and implements deterministic sampling and shuffling that are implemented in straightforward ways
+/// that will work identically across platforms. If you provide the same seed to the PRNG, you'll get the same results.
+/// This can be useful for cases like consensus applications, when a set of clients need to agree on pseudorandom
+/// results.
+///
+/// To avoid certain pitfalls, you need to choose a cryptographically-secure pseudorandom number generator (CSPRNG) that
+/// implements `CryptoRng + RngCore + SeedableRng`; a good choice is something like `ChaCha12Rng` from the `rand_chacha`
+/// crate (which happens to be the current `rand` default).
+///
+/// Once you instantiate the `DeterministicRandomizer` with your CSPRNG and a starting seed, you can start generating
+/// bounded pseudorandom numbers or sampling or shuffling arrays deterministically. You can always reseed the randomizer
+/// with a different seed, or reset it to the initial seed.
+///
+/// ```edition2018
+/// # use tari_crypto::deterministic_randomizer::DeterministicRandomizer;
+/// use rand_chacha::ChaCha12Rng;
+///
+/// // In consensus-type applications, the seed might come from a hash output whose input is some kind of state
+/// // We'll just make one arbitrarily; it must be `[u8; N]` with `N <= 32`
+/// let seed = [1u8; 32];
+/// let mut randomizer = DeterministicRandomizer::<ChaCha12Rng>::new(seed);
+///
+/// // Generate bounded numbers in the range `[0, 1000)`
+/// // Because it's deterministic, we know exactly what the results should be!
+/// assert_eq!(randomizer.next_bounded_u64(1000).unwrap(), 573);
+/// assert_eq!(randomizer.next_bounded_u64(1000).unwrap(), 786);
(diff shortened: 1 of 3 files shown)


hansieodendaal:
Really interesting, just wondering why this is not also implemented to produce a `u64` output, by reducing a `u128`.

AaronFeickert:
> Really interesting, just wondering why this is not also implemented (_or should I say required_) to produce a `u64` output, by reducing a `u128`.

This is just an initial example, so I didn't implement that (or any PRNG-based operations that might be needed). But agreed, that's one safe way to produce a `u64` value.

AaronFeickert:
Moving this from draft to ready, in case there's a desire to implement it directly.

src/deterministic_randomizer.rs
```diff
@@ -0,0 +1,114 @@
...
+    prng: R,
+    seed: <R as SeedableRng>::Seed,
+}
+
+impl<R> DeterministicRandomizer<R>
+where
+    R: CryptoRng + RngCore + SeedableRng,
+    <R as SeedableRng>::Seed: Clone,
+{
+    /// Initialize the randomizer with a seed
```
CjS77, line 32: An example here would be:fire:

src/deterministic_randomizer.rs
```diff
@@ -0,0 +1,114 @@
...
+
+    /// Reset the randomizer to its original seed
+    pub fn reset(&mut self) {
+        self.reseed(self.seed.clone());
+    }
+
+    /// Shuffle an array in place
+    pub fn shuffle<T>(&mut self, data: &mut [T]) -> Result<(), RandomizerError> {
+        // Restrict the array size to avoid bias
+        let checked_length = u32::try_from(data.len()).map_err(|_| RandomizerError)?;
```
CjS77, line 54: Why `u32` here?

AaronFeickert: One reason was that the original draft didn't support `u64` because of underlying PRNG API limits, which isn't the case anymore. The other reason was an assumption that the library should support 32-bit architectures, which (after discussion) is not a requirement.

Changed to assume 64-bit indexing.

CjS77:
I'm leaning towards all or nothing on the rand dependency - either impl our own `next_u64` (or `next_u128` since it's more apropos) or use rand functions throughout.

AaronFeickert:
> I'm leaning towards all or nothing on the rand dependency - either impl our own `next_u64` (or `next_u128` since it's more apropos) or use rand functions throughout.

The `next_u64` functionality is provided by `BlockRng` in `rand_core`, so it's not strictly a `rand` dependency. The only things we would need to dip into `rand` for are using such a `u64` in possibly version-specific modular reduction or shuffling routines, which is why we do those ourselves in this PR.

Do you think it could be the case that we'd still end up in dependency mismatch land if we update `rand` in the future and need to update `rand_core` with breaking changes as well?

CjS77:
> > I'm leaning towards all or nothing on the rand dependency - either impl our own `next_u64` (or `next_u128` since it's more apropos) or use rand functions throughout.
> 
> The `next_u64` functionality is provided by `BlockRng` in `rand_core`, so it's not strictly a `rand` dependency. The only things we would need to dip into `rand` for are using such a `u64` in possibly version-specific modular reduction or shuffling routines, which is why we do those ourselves in this PR.
> 
> Do you think it could be the case that we'd still end up in dependency mismatch land if we update `rand` in the future and need to update `rand_core` with breaking changes as well?

That is my concern, yes.

src/deterministic_randomizer.rs
```diff
@@ -0,0 +1,146 @@
...
+/// shuffling that can be used with PRNGs, there is no guarantee that these functions will always be implemented in the
+/// same way across versions, which could lead to incompatibility or dependency troubles. This library has no specific
+/// `rand` dependency, and implements deterministic bounded number generation and shuffling that are implemented in
+/// straightforward ways that will work identically across platforms. If you provide the same seed to the PRNG, you'll
+/// get the same results. This can be useful for cases like consensus applications, when a set of clients need to agree
+/// on pseudorandom results.
+///
+/// To avoid certain pitfalls, you need to choose a cryptographically-secure pseudorandom number generator (CSPRNG) that
+/// implements `CryptoRng + RngCore + SeedableRng`; a good choice is something like `ChaCha12Rng` from the `rand_chacha`
+/// crate (which happens to be the ) Once you instantiate the `DeterministicRandomizer` with your CSPRNG and a starting
```
hansieodendaal, line 27:
```suggestion
/// crate (which happens to be the ???). Once you instantiate the `DeterministicRandomizer` with your CSPRNG and a starting
```

hansieodendaal (dismissed):
utACK

src/deterministic_randomizer.rs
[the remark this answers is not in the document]
AaronFeickert: Yep, that's what it originally said. Must have been mangled during a rebase or something!

AaronFeickert:
@CjS77 brought up the idea that if we take this approach, it's better to implement `next_u32` and `next_u64` directly, rather than rely on the default implementations provided by `rand_core`.

src/deterministic_randomizer.rs
```diff
@@ -0,0 +1,188 @@
...
+        }
+    }
+
+    /// Reseed the randomizer with a new seed
+    pub fn reseed(&mut self, seed: <R as SeedableRng>::Seed) {
+        self.prng = R::from_seed(seed.clone());
+        self.seed = seed;
+    }
+
+    /// Reset the randomizer to its original seed
```
CjS77, line 100:
```suggestion
    /// Reset the randomizer with its current seed.
```

Since `r.seed(n); r.reseed(m); r.reset()` will use m and not n

CjS77 (approved):
One nit in the doc string, but LGTM

src/deterministic_randomizer.rs
AaronFeickert: Done!

CjS77 (approved):

PR #138, opened 2022-10-25
chore: fix various clippy errors

stringhandler:
There was an error in the clippy GHA workflow which prevented it from running. This has been fixed, but now there are a number of clippy errors. 

This PR fixes them

PR #141, opened 2022-10-27
fix: make the formatter happy

AaronFeickert:
Formatting and lint fixes so CI can pass. Gotta get that green check mark.

### DIFF
--- a/src/ffi/keys.rs
+++ b/src/ffi/keys.rs
@@ -17,8 +17,8 @@ use crate::{
     keys::{PublicKey, SecretKey},
     ristretto::{
         pedersen::commitment_factory::PedersenCommitmentFactory,
-        RistrettoComSig,
         RistrettoComAndPubSig,
+        RistrettoComSig,
         RistrettoPublicKey,
         RistrettoSchnorr,
         RistrettoSecretKey,
--- a/src/ffi/mod.rs
+++ b/src/ffi/mod.rs
@@ -9,7 +9,16 @@ mod error;
 mod keys;
 
 pub use error::lookup_error_message;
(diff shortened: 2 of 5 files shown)


AaronFeickert:
Looks like `clippy` is very unhappy with variables outside of format strings. Seems unhelpful.

sdbondi:
Looks like a change in rust, https://github.com/tari-project/tari-crypto/pull/138 uses latest nightly so that is probably why there is a sudden change in clippy

delta1:
Probably worth pinning the nightly version for fmt so you can upgrade it only when you want to

CjS77:
It's a really good lint, actually.
But better to make it a WARNING in `lints.toml` rather than hold the compiler version back. Then someone can come clean up after us.

delta1:
like @sdbondi said it just needs to be `allow`ed in lints.toml (it's already just a pedantic warning and cargo lint is defaulting to deny)

sdbondi:
> But better to make it a WARNING in lints.toml rather than hold the compiler version back. Then someone can come clean up after us.

I agree, might be a problem that you require +nightly to run cargo lints, that is `cargo +stable lints clippy` won't work until stable is on rust 1.65 (`uninlined_format_args` isn't recignised by clippy < 1.65), maybe thats ok. Otherwise, we should probably fix these lints.

CjS77:
> I agree, might be a problem that you require +nightly to run cargo lints, that is `cargo +stable lints clippy` won't work until stable is on rust 1.65 (`uninlined_format_args` isn't recignised by clippy < 1.65), maybe thats ok. Otherwise, we should probably fix these lints.

Yes, just ran into this on #143, but I think we can live with this until these features are stabilised

CjS77:
#143 covers this plus a few other bits and bobs, incl updating the code coverage script, so going to close this.

PR #143, opened 2022-11-02
chore: update tooling for stable only

CjS77:
The rust-toolchain.toml file has enforced Rust stable since July 2022,
but a lot of the tooling is still using nightly.

Partly this was due to necessity. Until recently, source code coverage
was a nightly only feature.

This PR updates the tooling to support stable where possible (we can't
kill nightly 100% yet).

* Update `test_coverage.sh` script to support changes in the API
* Only run CI flows on stable, where possible.
For example, clippy and cargo lints still need nightly for some of the
advanced configuration we use.


CjS77:
ACK

PR #145, opened 2022-11-03
feat!: improve signature api 

CjS77:
There are some footguns in the `SchnorrSignature::sign` method that this
PR seeks to mitigate.

Firstly, it's not clear in the function docs that the nonce and pubkey
are assumed to have been bound by the caller in the challenge.

The docs are updated to make this clearer.

Secondly, we deprecate the sign method and the utility module's sign
method in favour of:

- `sign_raw`, which is identical to the current sign method, but with a
  name that conveys the risk associated with using it.
- `sign_message`, which does what many clients might have thought `sign`
  does, which correctly binds a nonce and public key to the message
being signed in the challenge construction.
- Matching `verify_*` methods.

Tests and docs have been updated to reflect the changes.

The WASM and FFI library now use the domain-separated hashing algorithm
to generate challenges for signatures.

### DIFF
--- a/benches/signatures.rs
+++ b/benches/signatures.rs
@@ -47,7 +47,7 @@ fn sign_message(c: &mut Criterion) {
         b.iter_batched(
             gen_keypair,
             |d| {
-                let _sig = RistrettoSchnorr::sign_message(d.k, d.m.to_vec()).unwrap();
+                let _sig = RistrettoSchnorr::sign_message(&d.k, d.m.to_vec()).unwrap();
             },
             BatchSize::SmallInput,
         );
@@ -61,7 +61,7 @@ fn verify_message(c: &mut Criterion) {
         b.iter_batched(
             || {
                 let d = gen_keypair();
-                let s = RistrettoSchnorr::sign_message(d.k.clone(), d.m.to_vec()).unwrap();
+                let s = RistrettoSchnorr::sign_message(&d.k, d.m.to_vec()).unwrap();
                 (d, s)
             },
             |(d, s)| assert!(s.verify_message(&d.p, d.m.as_bytes())),
--- a/benches/signatures.rs
+++ b/benches/signatures.rs
@@ -9,7 +9,6 @@ use tari_crypto::{
     keys::{PublicKey, SecretKey},
     ristretto::{RistrettoPublicKey, RistrettoSchnorr, RistrettoSecretKey},
 };
-use tari_utilities::byte_array::ByteArray;
 
 fn generate_secret_key(c: &mut Criterion) {
     c.bench_function("Generate secret key", |b| {
@@ -46,7 +45,7 @@ fn sign_message(c: &mut Criterion) {
         b.iter_batched(
             gen_keypair,
             |d| {
-                let _sig = RistrettoSchnorr::sign_message(&d.k, &d.m).unwrap();
+                let _sig = RistrettoSchnorr::sign_message(&d.k, d.m).unwrap();
             },
             BatchSize::SmallInput,
         );
@@ -60,10 +59,10 @@ fn verify_message(c: &mut Criterion) {
         b.iter_batched(
             || {
                 let d = gen_keypair();
-                let s = RistrettoSchnorr::sign_message(&d.k, &d.m).unwrap();
+                let s = RistrettoSchnorr::sign_message(&d.k, d.m).unwrap();
                 (d, s)
             },
-            |(d, s)| assert!(s.verify_message(&d.p, &d.m)),
+            |(d, s)| assert!(s.verify_message(&d.p, d.m)),
             BatchSize::SmallInput,
         );
     });
--- a/benches/signatures.rs
+++ b/benches/signatures.rs
@@ -30,15 +30,14 @@ fn native_keypair(c: &mut Criterion) {
 struct SigningData {
     k: RistrettoSecretKey,
     p: RistrettoPublicKey,
-    m: RistrettoSecretKey,
+    m: [u8; 32],
 }
 
 fn gen_keypair() -> SigningData {
     let mut rng = thread_rng();
-    let mut msg = [0u8; 32];
-    rng.fill_bytes(&mut msg);
+    let mut m = [0u8; 32];
+    rng.fill_bytes(&mut m);
     let (k, p) = RistrettoPublicKey::random_keypair(&mut rng);
-    let m = RistrettoSecretKey::from_bytes(&msg).unwrap();
     SigningData { k, p, m }
 }
 
@@ -47,7 +46,7 @@ fn sign_message(c: &mut Criterion) {
         b.iter_batched(
             gen_keypair,
             |d| {
-                let _sig = RistrettoSchnorr::sign_message(&d.k, d.m.to_vec()).unwrap();
+                let _sig = RistrettoSchnorr::sign_message(&d.k, &d.m).unwrap();
             },
             BatchSize::SmallInput,
         );
@@ -61,10 +60,10 @@ fn verify_message(c: &mut Criterion) {
         b.iter_batched(
             || {
                 let d = gen_keypair();
-                let s = RistrettoSchnorr::sign_message(&d.k, d.m.to_vec()).unwrap();
+                let s = RistrettoSchnorr::sign_message(&d.k, &d.m).unwrap();
                 (d, s)
             },
-            |(d, s)| assert!(s.verify_message(&d.p, d.m.as_bytes())),
+            |(d, s)| assert!(s.verify_message(&d.p, &d.m)),
             BatchSize::SmallInput,
         );
     });
--- a/benches/signatures.rs
+++ b/benches/signatures.rs
@@ -30,7 +30,6 @@ fn native_keypair(c: &mut Criterion) {
 struct SigningData {
     k: RistrettoSecretKey,
     p: RistrettoPublicKey,
-    r: RistrettoSecretKey,
     m: RistrettoSecretKey,
 }
 
@@ -39,9 +38,8 @@ fn gen_keypair() -> SigningData {
     let mut msg = [0u8; 32];
     rng.fill_bytes(&mut msg);
     let (k, p) = RistrettoPublicKey::random_keypair(&mut rng);
-    let r = RistrettoSecretKey::random(&mut rng);
     let m = RistrettoSecretKey::from_bytes(&msg).unwrap();
-    SigningData { k, p, r, m }
+    SigningData { k, p, m }
 }
 
 fn sign_message(c: &mut Criterion) {
@@ -49,7 +47,7 @@ fn sign_message(c: &mut Criterion) {
         b.iter_batched(
             gen_keypair,
             |d| {
-                let _sig = RistrettoSchnorr::sign(d.k, d.r, &d.m.to_vec()).unwrap();
+                let _sig = RistrettoSchnorr::sign_message(d.k, d.m.to_vec()).unwrap();
             },
             BatchSize::SmallInput,
         );
@@ -63,10 +61,10 @@ fn verify_message(c: &mut Criterion) {
         b.iter_batched(
             || {
                 let d = gen_keypair();
-                let s = RistrettoSchnorr::sign(d.k.clone(), d.r.clone(), &d.m.to_vec()).unwrap();
+                let s = RistrettoSchnorr::sign_message(d.k.clone(), d.m.to_vec()).unwrap();
                 (d, s)
             },
-            |(d, s)| assert!(s.verify(&d.p, &d.m)),
+            |(d, s)| assert!(s.verify_message(&d.p, d.m.as_bytes())),
             BatchSize::SmallInput,
         );
     });
--- a/src/ffi/keys.rs
+++ b/src/ffi/keys.rs
@@ -23,6 +23,7 @@ use crate::{
         RistrettoSchnorr,
         RistrettoSecretKey,
     },
+    signatures::SchnorrSignature,
 };
 
 pub const KEY_LENGTH: usize = 32;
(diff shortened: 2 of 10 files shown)


src/ristretto/ristretto_sig.rs
```diff
@@ -167,11 +165,45 @@ mod test {
...
+    fn domain_separated_challenge() {
+        let P =
+            RistrettoPublicKey::from_hex("74896a30c89186b8194e25f8c1382f8d3081c5a182fb8f8a6d34f27fbefbfc70").unwrap();
+        let R =
+            RistrettoPublicKey::from_hex("fa14cb581ce5717248444721242e6b195a482d503a853dea4acb513074d8d803").unwrap();
+        let msg = "Moving Pictures";
+        let hash = SchnorrSignature::construct_domain_separated_challenge::<_, Blake256>(&R, &P, msg);
+        let naiive = Blake256::new()
+            .chain(R.as_bytes())
+            .chain(P.as_bytes())
```
AaronFeickert, line 188: If the other comment about swapping these two challenge inputs is accepted, there should be a corresponding swap here.

src/signatures/schnorr.rs
```diff
@@ -70,6 +107,63 @@ where
...
+    /// yourself, or even use [`sign_raw`] using a completely custom challenge.
+    pub fn sign_message<'a, B>(secret: &'a K, message: B) -> Result<Self, SchnorrSignatureError>
+    where
+        K: Add<Output = K> + Mul<&'a K, Output = K>,
+        B: AsRef<[u8]>,
+    {
+        let nonce = K::random(&mut rand::thread_rng());
+        let public_nonce = P::from_secret_key(&nonce);
+        let public_key = P::from_secret_key(secret);
+        let challenge = Self::construct_domain_separated_challenge::<_, Blake256>(&public_nonce, &public_key, message);
```
AaronFeickert, line 125: See other comment about swapping the first two challenge inputs.

src/wasm/key_utils.rs
```diff
@@ -156,18 +167,23 @@ pub(super) fn sign_message_with_key(
...
+    result: &mut SignResult,
+) {
     let (r, R) = match r {
         Some(r) => (r.clone(), RistrettoPublicKey::from_secret_key(r)),
         None => RistrettoPublicKey::random_keypair(&mut OsRng),
     };
-
-    let sig = match RistrettoSchnorr::sign(k.clone(), r, e) {
+    let P = RistrettoPublicKey::from_secret_key(k);
+    let e = SchnorrSignature::construct_domain_separated_challenge::<_, Blake256>(&R, &P, msg);
```
AaronFeickert, line 185: See other comment about swapping the first two challenge inputs.

src/signatures/schnorr.rs
```diff
@@ -70,6 +107,63 @@ where
...
+    /// message was signed by the secret key corresponding to the given public key, and that the challenge was
+    /// constructed using the domain-separation method defined in [`construct_domain_separated_challenge`].
+    pub fn verify_message<'a, B>(&self, public_key: &'a P, message: B) -> bool
+    where
+        for<'b> &'b K: Mul<&'a P, Output = P>,
+        for<'b> &'b P: Add<P, Output = P>,
+        B: AsRef<[u8]>,
+    {
+        let challenge =
+            Self::construct_domain_separated_challenge::<_, Blake256>(&self.public_nonce, public_key, message);
```
AaronFeickert, line 163: See other comment about swapping the first two challenge inputs.

src/signatures/schnorr.rs
```diff
@@ -57,8 +66,36 @@ where
...
+                instead, depending on your use case. This function will be removed in v1.0.0"
+    )]
+    #[allow(clippy::needless_pass_by_value)]
     pub fn sign(secret: K, nonce: K, challenge: &[u8]) -> Result<Self, SchnorrSignatureError>
-    where K: Add<Output = K> + Mul<P, Output = P> + Mul<Output = K> {
+    where
+        K: Add<Output = K>,
+        for<'a> K: Mul<&'a K, Output = K>,
+    {
+        Self::sign_raw(&secret, nonce, challenge)
```
AaronFeickert, line 87: Since the secret key is passed by value here, we could change the trait bound on `K` to require that it implement `Zeroize` (or modify `SecretKey` to require `Zeroize` instead), and then call `secret.zeroize()` before returning the signature. Then this library would never keep secret keys in memory.

I think a more comprehensive approach is to require that `SecretKey` implement `ZeroizeOnDrop` (if/when that marker trait becomes available to this crate), at which point you wouldn't need an explicit call to `zeroize()`, and could also provide better assurances that an implementer is handling secret keys better.

Right now it just so happens that the only implementer in this crate uses `RistrettoSecretKey`, which does zeroize on drop... but that's not a particular trait bound.

AaronFeickert (changes requested):

src/signatures/schnorr.rs
AaronFeickert: [This issue](https://github.com/tari-project/tari-crypto/issues/147) has been opened for discussion on the blockers to adding a `ZeroizeOnDrop` bound to `SecretKey`.

src/signatures/schnorr.rs
```diff
@@ -70,6 +107,63 @@ where
...
     }
 
+    /// Signs a message with the given secret key.
+    ///
+    /// This method correctly binds a nonce and the public key to the signature challenge, using domain-separated
+    /// hashing. The hasher is also opinionated in the sense that Blake2b 256-bit digest is always used.
+    ///
+    /// it is possible to customise the challenge by using [`construct_domain_separated_challenge`] and [`sign_raw`]
+    /// yourself, or even use [`sign_raw`] using a completely custom challenge.
+    pub fn sign_message<'a, B>(secret: &'a K, message: B) -> Result<Self, SchnorrSignatureError>
```
stringhandler, line 117: Since aggregate signatures like the kernel excess need to commit to the public nonces beforehand, the API would be more useful if the public nonce is passed in. 

`wasm32-unknown-unknown` also can't use any random functions, so this method cannot be compiled to it. Passing in the public nonce would help in this case too

src/signatures/schnorr.rs
```diff
@@ -10,11 +10,20 @@ use std::{
...
-use crate::keys::{PublicKey, SecretKey};
+use crate::{
+    hash::blake2::Blake256,
+    hash_domain,
+    hashing::{DomainSeparatedHash, DomainSeparatedHasher},
+    keys::{PublicKey, SecretKey},
+};
+
+// Define the hashing domain for Schnorr signatures
+hash_domain!(SchnorrSigChallenge, "com.tari.schnorr_signature", 1);
```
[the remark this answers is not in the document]
stringhandler: Agree here. I think we should rather pass in a domain to the signing function

src/signatures/schnorr.rs
AaronFeickert: So this means any application built on that target is responsible for handling its own randomness externally in general?

stringhandler: Yes. So in the case of `wasm32-unknown-unknown`, the WASM VM has no RNG. You can import a native function to provide it, depending on how you are hosting the VM

AaronFeickert: Would it then make sense to have this function also take an `Option<K>`, so the caller can provide its own nonce if required/desired? Or is that kind of fine-grained control over targets not possible within a function?

benches/signatures.rs
```diff
@@ -39,17 +38,16 @@ fn gen_keypair() -> SigningData {
     let mut msg = [0u8; 32];
     rng.fill_bytes(&mut msg);
     let (k, p) = RistrettoPublicKey::random_keypair(&mut rng);
-    let r = RistrettoSecretKey::random(&mut rng);
     let m = RistrettoSecretKey::from_bytes(&msg).unwrap();
```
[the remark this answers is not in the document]
CjS77: Likely due to the benches changing over time.

It doesn't affect the timings, since the secret key instantiation is done in the setup step. 

Easy enough to change.

src/signatures/schnorr.rs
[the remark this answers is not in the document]
CjS77: We definitely don't want to require a hash domain passed in all cases. There needs to be a minimal effort function call in this library. 
If someone wants to use tari_crypto to sign messages in the easiest possible way, then they should use this method.

The approach I landed on is  by including a 3rd generic type to the SchnorrSignature
struct definition.

The provision is optional. By default, the domain hash separator uses
`SchnorrSigChallenge` as the domain separation tag.

The `RsitrettoSchnorr` type alias is updated to include the default
domain separation tag, to keep backward compatibility. To provide a
custom hasher, update usages of `RistrettoSchnorr` t
`RistrettoSchnorrWithDomain`

src/signatures/schnorr.rs
```diff
@@ -70,6 +103,63 @@ where
...
+        public_key: &P,
+        message: B,
+    ) -> DomainSeparatedHash<D>
+    where
+        B: AsRef<[u8]>,
+        D: Digest,
+    {
+        DomainSeparatedHasher::<D, SchnorrSigChallenge>::new_with_label("challenge")
+            .chain(public_nonce.as_bytes())
+            .chain(public_key.as_bytes())
```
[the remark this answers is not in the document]
CjS77: You mean swapping the argument order? 

There can't be implementation confusion, since this is the implementation.

src/ffi/keys.rs
```diff
@@ -70,13 +71,15 @@ pub unsafe extern "C" fn sign(
...
+    let pubkey = RistrettoPublicKey::from_secret_key(&k);
     let r = RistrettoSecretKey::random(&mut OsRng);
+    let pub_r = RistrettoPublicKey::from_secret_key(&r);
     let msg = match CStr::from_ptr(msg).to_str() {
         Ok(s) => s,
         _ => return STR_CONV_ERR,
     };
-    let challenge = Blake256::digest(msg.as_bytes());
-    let sig = match RistrettoSchnorr::sign(k, r, &challenge) {
+    let e = SchnorrSignature::construct_domain_separated_challenge::<_, Blake256>(&pub_r, &pubkey, msg.as_bytes());
```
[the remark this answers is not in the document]
CjS77: The caller does not provide a nonce. The nonce argument is mutated to receive the public nonce value. I updated the docs to point this out and changed the argument name to be more specific.

src/ffi/keys.rs
```diff
@@ -123,13 +126,9 @@ pub unsafe extern "C" fn verify(
...
     };
+
     let sig = RistrettoSchnorr::new(r_pub, sig);
-    let challenge = Blake256::digest(msg.as_bytes());
-    let challenge = match RistrettoSecretKey::from_bytes(challenge.as_slice()) {
-        Ok(e) => e,
-        _ => return false,
-    };
-    sig.verify(&pk, &challenge)
+    sig.verify_message(&pk, msg.as_bytes())
```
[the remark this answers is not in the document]
CjS77: Changes challenge -> msg, and changed argument `*mut` to `*const` since the verify function takes 4 inputs.

src/signatures/schnorr.rs
CjS77: Zeroize trait bounds are a good idea, but out of scope for this PR.

src/signatures/schnorr.rs
CjS77: tari_crypto uses `getrandom/js` to provide a RNG inside wasm, so these methods do work.

In any event, `pub fn sign_with_nonce_and_message` is what you're looking for.

src/signatures/schnorr.rs
CjS77: We've pretty much standardised on this order. Swapping them at this stage produces greater risk of bugs.

src/signatures/schnorr.rs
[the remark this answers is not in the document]
AaronFeickert: I was thinking of the case where the developer uses the public `construct_domain_separated_challenge` as part of a custom use case. But on further thought that wouldn't really make sense, since you want to either use the safe wrapper functions (meaning you don't ever call this function) or handle the challenge yourself for a more complex use case (meaning you still wouldn't call this function).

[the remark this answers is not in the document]
CjS77: Yes, that's my thinking too.

FWIW, I was actually very torn about making `construct_domain_separated_challenge` public at all, but there might be times when you want to see how the sausage is made here, so I fell on the side of making it public,

benches/signatures.rs
```diff
@@ -30,26 +29,23 @@ fn native_keypair(c: &mut Criterion) {
 struct SigningData {
     k: RistrettoSecretKey,
     p: RistrettoPublicKey,
-    r: RistrettoSecretKey,
-    m: RistrettoSecretKey,
+    m: [u8; 32],
```
jorgeantonio21, line 32: From the current discussions around sensitive data being 'hidden' and/or zeroize. Shouldn't we heap allocate the data, using `Box` and wrap it around something like the `Hidden` type?

src/ristretto/ristretto_sig.rs
```diff
@@ -72,34 +71,69 @@ use crate::{
...
-/// let e = Blake256::digest(b"Maskerade");
-/// assert!(sig.verify_challenge(&P, &e));
+/// # #[allow(non_snake_case)]
+/// let P = RistrettoPublicKey::from_secret_key(&k);
+/// let sig: SchnorrSignature<RistrettoPublicKey, RistrettoSecretKey, MyCustomDomain> =
+///     SchnorrSignature::sign_message(&k, msg).unwrap();
+/// assert!(sig.verify_message(&P, msg));
 /// ```
-pub type RistrettoSchnorr = SchnorrSignature<RistrettoPublicKey, RistrettoSecretKey>;
+pub type RistrettoSchnorrWithDomain<H> = SchnorrSignature<RistrettoPublicKey, RistrettoSecretKey, H>;
```
jorgeantonio21, line 116: Neat :)

benches/signatures.rs
CjS77: Does not matter for tests or benchmarks.

AaronFeickert (approved):

CjS77:
ACK

Issue #147, opened 2022-11-03
Enforce zeroizing on generic secret keys


AaronFeickert:
Earlier work in [PR 137](https://github.com/tari-project/tari-crypto/pull/137) improves zeroizing support in the `RistrettoSecretKey` and `RistrettoPublicKey` types. However, neither of the generic `SecretKey` and `PublicKey` traits require zeroizing or zeroizing-on-drop support.

It may be useful to add trait bounds for these, in order to enforce better key hygiene at the trait level. We can already trivially add a `Zeroize` bound to both. However, much more useful would be to add a `ZeroizeOnDrop` bound to `SecretKey`, which requires that any implementation zeroize secret keys on drop. This can be especially useful in conjunction with other constructions, like signatures, that are defined generically.

The problem is that the current `zeroize` version does not support the `ZeroizeOnDrop` marker trait, instead only offering a custom macro for this purpose. The options are to either add our own marker trait and manually implement it for the Ristretto key wrappers (and any future key types), or update `zeroize` across the board and then replace the zeroize-on-drop macro with the more current `#[derive(ZeroizeOnDrop)]` macro. Updating to a "proper" version would conflict with the curve library. However, manual testing shows that a `zeroize` update in the curve library introduces no test failures (nor any failures on `bulletproofs` or `bulletproofs-plus`, which are dependencies of this crate).

I think this is another point in favor of doing a custom curve library fork. It seems unlikely that upstream curve library PRs will be accepted. Maintaining a fork seems straightforward, as upstream is infrequently updated. Further, this would enable functionality like [precomputation support](https://github.com/tari-project/bulletproofs-plus/issues/18) in `bulletproofs-plus`. The changes introduced by both of these features are minimal, suggesting little to no effort for ongoing maintenance; and no changes are breaking in a meaningful way that would prevent later reverting back to upstream.

PR #148, opened 2022-11-10
Functionality to work with `Hidden` types

AaronFeickert:
Work in `tari_utilities` [PR 52](https://github.com/tari-project/tari_utilities/pull/52) provides a new `Hidden` type for hidden data that supports safer handling of underlying data that supports the `Zeroize` trait. This is intended to handle things like passphrases, seed words, and derived key materials.

Initial work by @jorgeantonio21 on migrating existing key derivation function (KDF) outputs to use `Hidden` found two issues pretty quickly:
- The generic `SecretKey` and `PublicKey` traits do not enforce `Zeroize`, even though the Ristretto-based implementation in this crate do.
- The hashing API keeps copies of hash output data in memory that is not zeroized.

The first issue is more of an annoyance, since it means a `Hidden` type containing a generic `SecretKey` or `PublicKey` needs to add additional `Zeroize` trait bounds everywhere. This PR adds the trait bound to both. Since `RistrettoSecretKey` and `RistrettoPublicKey` already implement `Zeroize`, this change is seamless. This addresses part of [issue 147](https://github.com/tari-project/tari-crypto/issues/147).

The second issue is trickier. While we can't control what goes on inside the state of any underlying hash function (e.g. `Blake256` in the `tari` crate), and while some hashing API use cases don't assume the output is sensitive, KDF use cases do. For these cases, we want to minimize non-zeroized copies of the output. This PR mitigates this by implementing `FixedOutput` for `DomainSeparatedHasher`, which adds in-place output support. The use of the required `finalize_into_reset` _might_ be helpful in this case, since it doesn't consume the hasher but resets its internal state (albeit in a way that isn't clear to me).

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -31,7 +31,7 @@
 use std::{marker::PhantomData, ops::Deref};
 
 use blake2::VarBlake2b;
-use digest::{Digest, Output, Update};
+use digest::{Digest, Output, Update, FixedOutput};
 use sha3::Sha3_256;
 use tari_utilities::ByteArray;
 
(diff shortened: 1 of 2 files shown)


AaronFeickert:
Still needs to be tested against the `tari` repository to ensure it doesn't open up any weird issues.

AaronFeickert:
This should be checked by @jorgeantonio21 to see if it properly addresses the issues he had.

src/hashing.rs
```diff
@@ -628,6 +642,26 @@ mod test {
...
     }
 
+    #[test]
+    fn finalize_into() {
+        hash_domain!(TestHasher, "com.example.test");
+        let mut hasher = DomainSeparatedHasher::<Blake256, TestHasher>::new();
+        hasher.update([0, 0, 0]);
+
+        let mut output = GenericArray::<u8, <Blake256 as Digest>::OutputSize>::default();
+        hasher.finalize_into(&mut output);
```
jorgeantonio21, line 652: Maybe add a new line with an assertion to compare the result of output with the same data, hashed using `finalize`.

src/hashing.rs
```diff
@@ -628,6 +642,26 @@ mod test {
...
+    }
+
+    #[test]
+    fn finalize_into_reset() {
+        hash_domain!(TestHasher, "com.example.test");
+        let mut hasher = DomainSeparatedHasher::<Blake256, TestHasher>::new();
+        hasher.update([0, 0, 0]);
+
+        let mut output = GenericArray::<u8, <Blake256 as Digest>::OutputSize>::default();
+        hasher.finalize_into_reset(&mut output);
```
jorgeantonio21, line 662: Same as above

jorgeantonio21 (approved):

CjS77 (approved):
+1

CjS77:
Will merge once there's an ACK

CjS77:
ACK

PR #149, opened 2022-11-14
fix!: signature serialization

AaronFeickert:
The recent [update](https://github.com/tari-project/tari-crypto/pull/145) to the Schnorr signature API introduced a breaking serialization change due to the presence of a new `PhantomData` field that serializes to a null value, which breaks compatibility with existing signatures.

This work introduces a simple change that instructs `serde` to ignore the field during serialization and deserialization. There shouldn't be any other problems with deserialization, since `serde` will apply the correct default type.

BREAKING CHANGE: Technically this is a breaking change to serialization, even though it restores the expected serialization for proper compatibility with other crates.

### DIFF
--- a/src/signatures/schnorr.rs
+++ b/src/signatures/schnorr.rs
@@ -45,6 +45,7 @@ pub enum SchnorrSignatureError {
 pub struct SchnorrSignature<P, K, H = SchnorrSigChallenge> {
     public_nonce: P,
     signature: K,
+    #[serde(skip)]
     _phantom: PhantomData<H>,
 }
 


AaronFeickert:
This is blocked on testing with the `tari` crate, which is where the deserialization test failures occur.

AaronFeickert:
The `tari` crate, which is where the deserialization break was observed, has all tests passing using this version of `tari-crypto` except one. The broken test appears to have been broken already.

CjS77 (approved):

PR #150, opened 2022-11-15
feat: add Borsh

Cifko:
Add Borsh(De)Serialize for keys and signatures.


CjS77:
> This also needs to be behind a `borsh` feature, as I mentioned in tari_utilities

+1 on this

CjS77 (approved):
lgtm

src/keys.rs
```diff
@@ -40,6 +42,49 @@ pub trait SecretKey: ByteArray + Clone + PartialEq + Eq + Add<Output = Self> + D
 /// implementations need to implement this trait for them to be used in Tari.
 ///
 /// See [SecretKey](trait.SecretKey.html) for an example.
+/// cbindgen:ignore
+#[cfg(feature = "borsh")]
+pub trait PublicKey:
```
sdbondi, line 47: I think having a different trait depending on whether the borsh feature flag is enabled or not could be troublesome.

Any reason we require these trait bounds in the trait itself?

stringhandler (approved):

PR #151, opened 2022-11-18
feat: update to the newer tag version of tari_utilities

jorgeantonio21:
Upgrading the tari utilities dependency tag version


sdbondi:
utACK

sdbondi:
Closed in #152

PR #153, opened 2022-11-21
chore: add partial com and pub sigs tests

SWvheerden:
Add better tests for `ristretto_com_and_pub_sig`
Also, prove that partial signing is possible.


CjS77 (approved):
lgtm

PR #160, opened 2022-12-13
fix: use updated range proof API

AaronFeickert:
Uses a new [range proof API](https://github.com/tari-project/bulletproofs-plus/pull/22).

### DIFF
--- a/src/ristretto/bulletproofs_plus.rs
+++ b/src/ristretto/bulletproofs_plus.rs
@@ -11,7 +11,7 @@ use bulletproofs_plus::{
     extended_mask::ExtendedMask as BulletproofsExtendedMask,
     generators::pedersen_gens::ExtensionDegree as BulletproofsExtensionDegree,
     range_parameters::RangeParameters,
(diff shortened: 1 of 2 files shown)


src/ristretto/bulletproofs_plus.rs
```diff
@@ -237,9 +237,12 @@ impl RangeProofService for BulletproofsPlusService {
                     minimum_value_promises: vec![None],
                     seed_nonce: None,
                 };
-                match RistrettoRangeProof::verify_do_not_recover_masks(self.transcript_label, &[statement], &[
-                    rp.clone()
-                ]) {
+                match RistrettoRangeProof::verify_batch(
```
stringhandler, line 240: I'm not a fan of returning different data based on the action, but I think that's subjective, so think it's fine

stringhandler (approved):
I generally prefer having two methods with different return types rather than one method that returns useful data sometimes, and sentinel values another, but it's subjective so not too fussed

PR #162, opened 2023-01-11
fix: check for valid `Blake2b` keys and test parameters

AaronFeickert:
When instantiating a `Blake2b` hasher, parameters need to be checked for validity. Currently, the salt, persona, and output lengths are checked to ensure they are valid. However, the key length is not checked, which can lead to a panic.

This PR checks for a valid key length (between 0 and 64 bytes). It also adds a test for each parameter to ensure proper failure.


CjS77 (approved):
LGTM

PR #167, opened 2023-02-22
feat: add missing methods to Commitment signature

jorgeantonio21:
Add [`from_vec`] and [`from_bytes`] to `CommitmentSignature` type. Also adds a `sign_raw` method for it.

### DIFF
--- a/src/signatures/commitment_signature.rs
+++ b/src/signatures/commitment_signature.rs
@@ -8,7 +8,7 @@ use std::{
 };
 
 use serde::{Deserialize, Serialize};
-use tari_utilities::ByteArray;
+use tari_utilities::{ByteArray, ByteArrayError};
 use thiserror::Error;
 
 use crate::{
@@ -66,7 +66,9 @@ where
 
     /// This is the left-hand side of the signature verification equation
     pub fn calc_signature_verifier<C>(&self, factory: &C) -> HomomorphicCommitment<P>
-    where C: HomomorphicCommitmentFactory<P = P> {
+    where
+        C: HomomorphicCommitmentFactory<P = P>,
+    {
         // v*H + u*G
         factory.commit(&self.u, &self.v)
     }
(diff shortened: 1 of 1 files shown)


src/signatures/commitment_signature.rs
```diff
@@ -159,14 +183,31 @@ where
         &self.public_nonce
     }
 
-    /// Returns a canonical byte representation of the commitment signature
+    /// Returns a canonical vector of bytes representation of the commitment signature
```
sdbondi, line 186: A vector of bytes would be `Vec<bytes>` where `bytes` is probably `Vec<u8>` :)

jorgeantonio21: Nice catch eheh :) did refactor accordingly

src/signatures/commitment_signature.rs
```diff
@@ -81,6 +83,28 @@ where
...
+    {
+        Self::sign_raw(secret_a, secret_x, nonce_a, nonce_x, challenge, factory)
+    }
+
+    /// Sign the provided challenge with the value commitment's value and blinding factor. The two nonces should be
+    /// completely random and never reused - that responsibility lies with the calling function.
+    ///
+    /// WARNING: The provided secret keys and nonces are NOT bound to the challenge. This method assumes that the
+    /// challenge has been constructed such that all commitments are already included in the challenge.
+    pub fn sign_raw<C>(
```
sdbondi, line 100: Why was this added? As far as I can see it's the same as `sign`

jorgeantonio21: Should we replace the `sign` function by this `sign_raw`, to convey the idea that we are making assumptions that the inputs are sensible ? Similarly to what was used in the `SchnorrSignature`, via the signature api ?

sdbondi: Ah right, I think this should be in another PR because we should also add a deprecation attribute to `sign` which may involve a number of other changes.

stringhandler (approved):
Happy. PR description must be updated

PR #168, opened 2023-02-25
test: add simple aggregated range proof test

hansieodendaal:
Added a simple aggregated range proof test to show the minimum constructs needed if we are only interested in an aggregated range proof, possibly with more than one commitment extension degree.

### DIFF
--- a/src/ristretto/bulletproofs_plus.rs
+++ b/src/ristretto/bulletproofs_plus.rs
@@ -656,7 +656,7 @@ mod test {
                 let mut commitment_value_map_private = HashMap::new();
 
                 #[allow(clippy::cast_possible_truncation)]
-                let (value_min, value_max) = (0u64, (1u128 << (bit_length - 1)) as u64);
+                let (value_min, value_max) = (0u64, ((1u128 << bit_length) - 1) as u64);
                 for aggregation_size in AGGREGATION_SIZE {
                     // 1. Prover's service
                     let bulletproofs_plus_service =
@@ -794,7 +794,7 @@ mod test {
                 let mut statements_public = vec![];
 
                 #[allow(clippy::cast_possible_truncation)]
-                let (value_min, value_max) = (0u64, (1u128 << (bit_length - 1)) as u64);
+                let (value_min, value_max) = (0u64, ((1u128 << bit_length) - 1) as u64);
 
                 // 1. Prover's service
                 let bulletproofs_plus_service =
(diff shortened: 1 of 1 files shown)


sdbondi (approved):
Thanks, nice and easy to understand

AaronFeickert (dismissed):

src/ristretto/bulletproofs_plus.rs
```diff
@@ -777,6 +777,72 @@ mod test {
...
+        ] {
+            let factory = ExtendedPedersenCommitmentFactory::new_with_extension_degree(extension_degree).unwrap();
+
+            for aggregation_size in [2, 4] {
+                // 0.  Batch data
+                let mut proofs = vec![];
+                let mut statements_public = vec![];
+
+                #[allow(clippy::cast_possible_truncation)]
+                let (value_min, value_max) = (0u64, (1u128 << (bit_length - 1)) as u64);
```
AaronFeickert, line 797: Nit: This isn't the full range of allowed values.

hansieodendaal:
```suggestion
                let (value_min, value_max) = (0u64, (1u128 << bit_length) - 1) as u64);
```

AaronFeickert: This same construction is also used elsewhere in the same test file: [here](https://github.com/tari-project/tari-crypto/blob/abbd7208a4fef9664d233aff50a34a9637825dbe/src/ristretto/bulletproofs_plus.rs#L659) and [here](https://github.com/tari-project/tari-crypto/blob/abbd7208a4fef9664d233aff50a34a9637825dbe/src/ristretto/bulletproofs_plus.rs#L788) and [here](https://github.com/tari-project/tari-crypto/blob/abbd7208a4fef9664d233aff50a34a9637825dbe/src/ristretto/bulletproofs_plus.rs#L895).

However, the subsequent calls to `gen_range` are exclusive of the high value, so this will never cover the entire value range. This isn't problematic (since these are just randomized tests), but interesting to note.

The change you suggest (applied to the other instances as well) seems useful, so it's more clear to the reader what the intended value range is.

hansieodendaal: Agreed, changed it all - good pickup, this was an oversight  :-)

AaronFeickert (approved):
LGTM

stringhandler (approved):

PR #171, opened 2023-03-14
feat: require secret keys to be zeroized on drop

AaronFeickert:
Adds a trait bound requiring that a `SecretKey` zeroize on drop. Derives this trait bound for `RistrettoSecretKey`.

Currently, we use a macro from `zeroize` to zeroize a `RistrettoSecretKey` on drop. This works fine, but doesn't let us require this behavior as a `SecretKey` trait bound. Now that we have a custom curve library fork that supports an updated version of `zeroize`, we can take advantage of a new `ZeroizeOnDrop` derive macro that adds a corresponding marker trait. This PR adds the marker trait as a trait bound to `SecretKey` and uses the derive macro for `RistrettoSecretKey`. This adds an additional layer of safety to secret keys.

Closes [issue #147](https://github.com/tari-project/tari-crypto/issues/147).

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -11,7 +11,7 @@ use std::ops::Add;
 use rand::{CryptoRng, Rng};
 use serde::{de::DeserializeOwned, ser::Serialize};
 use tari_utilities::ByteArray;
-use zeroize::Zeroize;
+use zeroize::{Zeroize, ZeroizeOnDrop};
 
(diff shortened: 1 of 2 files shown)


AaronFeickert:
This should be tested on the main `tari` repository to ensure it doesn't introduce any dependency issues.

stringhandler (approved):

PR #173, opened 2023-03-20
feat: `RistrettoSecretKey` inversion

AaronFeickert:
Adds multiplicative inversion to `RistrettoSecretKey`. Adds a few unit tests.

Note that the underlying `invert` functionality from `curve25519-dalek` does not assert that the `Scalar` in question is nonzero, but requires the caller to check this; inverting zero is undefined. This PR specifically checks this, and as a result, the inversion returns an `Option`. This is probably annoying for the caller, but might stop bad things from happening.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -34,6 +34,8 @@ pub trait SecretKey:
     fn key_length() -> usize;
     /// Generates a random secret key
     fn random<R: Rng + CryptoRng>(rng: &mut R) -> Self;
+    /// Get the multiplicative inverse of a nonzero secret key
+    fn invert(&self) -> Option<Self>;
 }
 
 //----------------------------------------   Public Keys  ----------------------------------------//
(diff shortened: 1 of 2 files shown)


hansieodendaal (approved):
utACK

CjS77:
What is the inversion of a secret key?

Other than a leaky abstraction, that is

AaronFeickert:
> What is the inversion of a secret key?
> 
> Other than a leaky abstraction, that is

From a general viewpoint, it's undefined. But so are things like addition and multiplication, which are `SecretKey` trait bounds too, since the protocol basically assumes secret keys are field elements.

AaronFeickert:
After discussion with @CjS77, inversion functionality has been moved out of `SecretKey` and is implemented directly for `RistrettoSecretKey` instead.

CjS77 (dismissed):
lgtm

hansieodendaal (approved):
utACK

stringhandler (approved):

PR #174, opened 2023-03-21
test: revert and freeze `wasm-pack` module

AaronFeickert:
There's an issue with WASM builds and tests on the current 0.11.0 version of the `wasm-pack` Node.js module that isn't present in the 0.10.3 version.

Until that issue is fixed upstream, this PR reverts to and freezes at the known working version.

### DIFF
--- a/.github/workflows/wasm.yml
+++ b/.github/workflows/wasm.yml
@@ -20,7 +20,7 @@ jobs:
           target: wasm32-unknown-unknown
           override: true
       - name: install wasm-pack
-        run: npm install -g wasm-pack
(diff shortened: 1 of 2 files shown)


AaronFeickert:
This is now an [upstream issue](https://github.com/rustwasm/wasm-pack/issues/1248).

AaronFeickert:
This now removes the `--enable-mutable-globals` flag on `wasm-opt`, which appears to be the cause of a new CI failure. From reading elsewhere, this flag should be enabled by default and therefore unnecessary anyway. However, I wasn't able to reproduce the failure locally.

stringhandler (approved):

PR #175, opened 2023-04-03
fix: dependency on multiple versions of rand_core

hansieodendaal:
Fixed dependency on multiple versions of `rand_core` by downgrading `rand_chacha` and the direct reference to `rand_core`.

This is important to mitigate dependency clashes in future.


SWvheerden (changes requested):
Why are we downgrading the rand core version?
Can we not update the other dependencies to get to rand_core 0.6?

hansieodendaal:
> Why are we downgrading the rand core version? Can we not update the other dependencies to get to rand_core 0.6?

The current thinking as I have it is to use the lowest common denominator everywhere.

PR #177, opened 2023-04-28
feat!: range proof API change

AaronFeickert:
A pending [pull request](https://github.com/tari-project/bulletproofs-plus/pull/29) in the `bulletproofs-plus` repository changes the range proof API to better support [proof offloading](https://github.com/tari-project/bulletproofs-plus/issues/27) more safely. In particular, it requires the use of two seed nonces for mask recovery, and changes these to be byte arrays in order to be more abstract to the underlying elliptic curve group, and more efficient.

This work makes an update to support the new API.

### DIFF
--- a/src/ristretto/bulletproofs_plus.rs
+++ b/src/ristretto/bulletproofs_plus.rs
@@ -12,7 +12,7 @@ use bulletproofs_plus::{
     generators::pedersen_gens::ExtensionDegree as BulletproofsExtensionDegree,
     range_parameters::RangeParameters,
     range_proof::{RangeProof, VerifyAction},
-    range_statement::RangeStatement,
+    range_statement::{RangeSeedNonce, RangeStatement},
     range_witness::RangeWitness,
     PedersenGens,
 };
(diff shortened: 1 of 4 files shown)


AaronFeickert:
Do not merge. This currently references a `bulletproofs-plus` feature branch dependency.

src/ristretto/bulletproofs_plus.rs
```diff
@@ -157,7 +156,7 @@ impl BulletproofsPlusService {
     /// Helper function to prepare a batch of private range statements
     pub fn prepare_private_range_statements(
         &self,
-        statements: Vec<&RistrettoAggregatedPrivateStatement>,
+        statements: &Vec<&RistrettoAggregatedPrivateStatement>,
```
sdbondi, line 159: nit: References to Vec force callers to use a Vec instead of anything that dereferences to a slice, with no benefit. https://rust-lang.github.io/rust-clippy/v0.0.212/index.html#ptr_arg

```suggestion
        statements: &[&RistrettoAggregatedPrivateStatement],
```

AaronFeickert: Yeah, good point. I'd made that fix elsewhere at the linter's suggestion, but must have missed it here. I'll take another look at the other associated PRs as well.

PR #181, opened 2023-06-05
feat!: expose secret key length as a constant

AaronFeickert:
Currently, the `PublicKey` trait exposes the encoded key byte length via the associated `KEY_LEN` constant, and also provides it via the `key_length` function. However, the `SecretKey` trait only exposes its encoded key byte length via the `key_length` function, which cannot be constant. There are use cases where it's handy to get the key size at compile time, particularly when you want to use arrays and avoid heap allocations.

This PR unifies the `PublicKey` and `SecretKey` traits by adding an associated `KEY_LEN` constant to `SecretKey`. It then updates the `RistrettoSecretKey` implementation accordingly. This retains all existing functionality, so the change is not breaking.

### DIFF
--- a/src/keys.rs
+++ b/src/keys.rs
@@ -30,8 +30,14 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 pub trait SecretKey:
     ByteArray + Clone + PartialEq + Eq + Add<Output = Self> + Default + Zeroize + ZeroizeOnDrop
 {
-    /// The length of the key, in bytes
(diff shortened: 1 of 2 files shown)


sdbondi (approved):
Thanks!

sdbondi:
Technically, this change is breaking in the sense that if anyone implemented the `PublicKey` and/or `SecretKey` traits, their code would not compile after this update. Of course, the impact on anyone is likely zero.

PR #183, opened 2023-07-07
feat!: add hardware wallet support

hansieodendaal:
Added hardware wallet support

- Updated `curve25519-dalek` to `version = "4.0.3"` with all related dependencies and updated the code to reflect all API changes.
- Added feature flags to use non-default features for most dependencies.
- Added CI checks to ensure hardware wallet support by building the library to `--target=thumbv8m.main-none-eabi`.
- Cleaned up CI in general as the previous exclusions are not necessary anymore.
- Updated the README.md

_**Note:** `thiserror` was dropped in favour of `snafu` as `thiserror` makes use of the standard library and cannot build to non-standard targets._

Co-Authored-By: SW van Heerden <swvheerden@gmail.com>


AaronFeickert:
A new version of the curve library [has been tagged](https://github.com/tari-project/curve25519-dalek/releases/tag/v4.0.3) and should presumably be used here.

Dependency updates have also been [merged](https://github.com/tari-project/bulletproofs-plus/pull/38) into `bulletproofs-plus` as well.

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
+once_cell = {version = "1.8", default-features = false }
+rand_chacha = { version    = "0.3", default-features = false }
+rand_core = {version = "0.6", default-features = false }
+serde = { version = "1.0", default-features = false, optional = true }
+serde_json = { version = "1.0", default-features = false, optional = true }
 serde-wasm-bindgen = { version = "0.4", optional = true }
-sha3 = "0.9.0"
-thiserror = "1.0.20"
+sha3 = {version = "0.10", default-features = false }
+snafu = {version = "0.7.4", default-features = false }
```
SWvheerden, line 32:
```suggestion
snafu = {version = "0.7", default-features = false }
```

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
-rand = { version = "0.7.3", default-features = false }
-rand_chacha = "0.3.1"
-rand_core = "0.6.4"
-serde = "1.0"
-serde_json = "1.0"
+tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", version = "0.5.0", default-features = false }
+blake2 = {version = "0.10", default-features = false }
+borsh = { version = "0.10", default-features = false, optional = true }
+bulletproofs = { package = "tari_bulletproofs", git = "https://github.com/tari-project/bulletproofs", version = "4.4.1", default-features = false, optional = true }
+bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", version = "0.3.0", default-features = false, optional = true }
```
SWvheerden, line 18:
```suggestion
bulletproofs_plus = { package = "tari_bulletproofs_plus", git = "https://github.com/tari-project/bulletproofs-plus", version = "0.3", default-features = false, optional = true }
```

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
-lazy_static = "1.3.0"
-log = "0.4.0"
-merlin = { version = "2.0.1", default-features = false }
-once_cell = "1.8.0"
-rand = { version = "0.7.3", default-features = false }
-rand_chacha = "0.3.1"
-rand_core = "0.6.4"
-serde = "1.0"
-serde_json = "1.0"
+tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", version = "0.5.0", default-features = false }
```
SWvheerden, line 14:
```suggestion
tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", version = "0.5", default-features = false }
```

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
-sha2 = "0.9.5"
-wasm-bindgen-test = "0.3.24"
+rand = "0.8"
+bincode = "1.1"
+criterion = {version = "0.5", default-features = false }
+rand_chacha = "0.3"
+sha2 = "0.10"
+wasm-bindgen-test = "0.3"
+zeroize = {version = "1", default-features = false }
+tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", version = "0.5.0", default-features = false, features = ["zero", "std"] }
```
SWvheerden, line 45:
```suggestion
tari_utilities = { git = "https://github.com/tari-project/tari_utilities.git", version = "0.5", default-features = false, features = ["zero", "std"] }
```

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
 
 [build-dependencies]
-cbindgen = "0.17.0"
+cbindgen = {version = "0.17.0", optional = true }
```
SWvheerden, line 48:
```suggestion
cbindgen = {version = "0.17", optional = true }
```

Cargo.toml
```diff
@@ -11,53 +11,60 @@ version = "0.17.0"
...
+borsh_ser = ["borsh"]
 
 [lib]
 # Disable benchmarks to allow Criterion to take over
 bench = false
-crate-type = ["lib", "cdylib", "staticlib"]
+# To build for "staticlib" use `cargo rustc --crate-type staticlib --release ...`
+crate-type = ["lib", "cdylib"] # ("staticlib" does not work with hardware wallet, "#[global_allocator]" issue)
+
+#[lib.profile.no_std]
```
SWvheerden, line 67:
```suggestion
```

src/hash/mod.rs
```diff
@@ -3,4 +3,4 @@
 
 //! Convenience wrappers for hash functions
 pub mod blake2;
-pub mod error;
+// pub mod error;
```
SWvheerden, line 6:
```suggestion
```

SWvheerden: and we need to remove error.rs here as well

SWvheerden (commented):
just a few small things, but other than that it looks good

SWvheerden (approved):

.github/workflows/clippy-check.yml
```diff
@@ -30,8 +37,19 @@ jobs:
...
+        with:
+          command: check
+          args: --release --all-targets --all-features
+      - name: Cargo check no default for non std
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --release --all-targets
+          args: --release --no-default-features --features "no_std" --features "borsh_ser"
+      - name: Cargo build no default for non std target
```
stringhandler, line 50: I think you need some comments explaining what these do, why it seems to be a duplicate of the above and when (or never) should they be removed

.github/workflows/test.yml
```diff
@@ -23,32 +23,24 @@ jobs:
...
         uses: actions-rs/cargo@v1
         with:
           command: check
-          args: --benches
+          args: --benches --all-features
+      - name: ledger no-std
+        uses: actions-rs/cargo@v1
+        with:
+          command: check
+          args: --release --features "no_std" --features "borsh_ser"
```
stringhandler, line 46: Are these two features dependent on each other? It should either be split into 2 tasks, or changed to be dependent on one another. Also you are missing --no-default-features.

In fact, this job is a repeat from the above file. Is this intended?

Cargo.toml
```diff
@@ -11,53 +11,58 @@ version = "0.17.0"
...
 [features]
-default = ["u64_backend"]
-u64_backend = ["curve25519-dalek/u64_backend"]
-simd_backend = ["curve25519-dalek/simd_backend", "bulletproofs/simd_backend"]
-wasm = ["wasm-bindgen", "getrandom/js", "rand/wasm-bindgen", "serde-wasm-bindgen"]
-ffi = []
+default = ["serialize", "bulletproofs", "bulletproofs_plus", "zero", "precomputed_tables", "borsh_ser", "borsh/std"]
+wasm = ["wasm-bindgen", "getrandom/js", "serde-wasm-bindgen", "rand"]
+ffi = ["cbindgen", "rand"]
+zero = ["zeroize", "tari_utilities/zero"]
```
stringhandler, line 54: I meant to actually raise this in the tari_utilities crate, but this feature would be better named `zeroize`.

Cargo.toml
```diff
@@ -11,53 +11,58 @@ version = "0.17.0"
...
 musig = []
+serialize = ["serde", "serde_json", "curve25519-dalek/serde", "tari_utilities/serialize"]
+no_std = ["lazy_static/spin_no_std"]
+precomputed_tables = []
+borsh_ser = ["borsh"]
```
stringhandler, line 59: why not just `borsh` ? Where does `ser` come from?

README.md
```diff
@@ -46,29 +73,74 @@ console.log("PB = ", keys.public_key("Bob"));
...
+
+This will include support for WebAssembly.
+
+### ffi
+
+This will include support for the C FFI module.
+
+### zero
+
+This will include support for zeroing memory.
```
stringhandler, line 134: I think this could be expanded and should add a warning about not enabling it. I would also mention that it uses the zeroize crate

README.md
```diff
@@ -46,29 +73,74 @@ console.log("PB = ", keys.public_key("Bob"));
...
+
+### precomputed_tables
+
+This will include support for precomputed tables for the curve.
+
+### musig
```
stringhandler, line 140: Is this still present? I thought we removed it

src/hash/blake2.rs
```diff
@@ -3,87 +3,25 @@
...
-    }
-}
-
-impl Update for Blake256 {
-    fn update(&mut self, data: impl AsRef<[u8]>) {
-        self.0.update(data);
-    }
-}
+/// Blake2b 256-bit hash function
+pub type Blake256 = Blake2b<U32>;
```
stringhandler, line 11: Perhaps its better just to delete this file entirely, I don't think it's used anymore

src/hashing.rs
```diff
@@ -303,43 +304,65 @@ pub trait AsFixedBytes<const I: usize>: AsRef<[u8]> {
...
+                provided: hash_vec_length,
+            });
         }
         let mut buffer: [u8; I] = [0; I];
         buffer.copy_from_slice(&hash_vec[..I]);
         Ok(buffer)
     }
 }
 
+impl<TInnerDigest: OutputSizeUser, TDomain: DomainSeparation> OutputSizeUser
```
stringhandler, line 318: These changes seem extraneous to the PR and should be put in a separate PR. It's hard to know what the purpose of these changes are

src/ristretto/bulletproofs_plus.rs
```diff
@@ -73,10 +74,10 @@ impl TryFrom<&BulletproofsExtendedMask> for RistrettoExtendedMask {
     fn try_from(extended_mask: &BulletproofsExtendedMask) -> Result<Self, Self::Error> {
         let secrets = extended_mask
             .blindings()
-            .map_err(|e| RangeProofError::ExtensionDegree(e.to_string()))?;
+            .map_err(|e| RangeProofError::RPExtensionDegree { reason: e.to_string() })?;
```
stringhandler, line 77: This abbreviation is not necessary. Rather keep it `ExtensionDegree`. Adding an abbreviation does little to help here

src/ristretto/ristretto_keys.rs
```diff
@@ -1021,6 +1031,8 @@ mod test {
 
     #[cfg(feature = "borsh")]
```
stringhandler, line 1032: This needs to be fixed if you are going to rename to `borsh_ser`

stringhandler (changes requested):
I cannot in good conscience allow this to be merged.

There is a mix of changes that arise from getting this to compile with newer versions of libraries and also large changes to the domain signatures.

The spattering of feature flags all over is prone to error. You missed at least one borsh rename and it is hard as a developer to see any kind of pattern if they are adding new code, especially if there is a feature flag. I would suggest trying to move logic that requires a feature flag into a separate file for the feature, e.g. serialize methods in a file called `keys` into a `keys_serialize` file.

In addtion:

1. I do not like the naming of Super in SuperSecretKey and SuperPublicKey
2. It seems unsafe to allow Zeroize to be disabled so easily. I do not like the zeroize feature at all, is there a reason for this?
3. I would prefer Bulletproofs to just be removed. The code will never be tested, yet we are making updates to it. Someone may use this library and incorrectly think that it is stable and tests. 
4. Borsh is only used for serialization, I don't understand why it should be renamed to `borsh_ser`
5. There are some features that are not tested with CI, and with so many different features, I would suggest adding a test for each feature individually to catch compilation errors

Please move the changes to digest and domain separation to a new PR.

Cargo.toml
SWvheerden, line 54:
```suggestion
zero = ["zeroize", "tari_utilities/zero", "curve25519-dalek/zeroize"]
```

src/deterministic_randomizer.rs
```diff
@@ -147,6 +151,18 @@ where
...
+where
+    R: CryptoRng + RngCore + SeedableRng,
+    <R as SeedableRng>::Seed: Clone,
+{
+    fn zeroize(&mut self) {
+        let seed = <R as SeedableRng>::Seed::default();
+        *self = DeterministicRandomizer::<R>::new(seed);
+    }
+}
+
```
AaronFeickert, line 165: Not a fan of this approach, since this is isn't true zeroization and doesn't provide the same guarantees against compiler optimization.

If you can implement `DefaultIsZeroes` on the `Seed` associated type, you could properly zeroize it. However, there's no guarantee that the PRNG state isn't leaky through the reset.

Overall, I think it's a bad idea to handle bespoke zeroization in this way, as it would provide the user with a sense of security that isn't supported by the implementation. I would remove this entirely.

src/dhke.rs
```diff
@@ -34,9 +35,9 @@ where
         self.0.as_bytes()
     }
 }
-
+#[cfg(feature = "zero")]
 impl<P> Zeroize for DiffieHellmanSharedSecret<P>
-where P: Zeroize
+where P: PublicKey
```
AaronFeickert, line 40: I'm probably missing something here, but how does this even compile if there's no `Zeroize` trait bound on `P` at any time?

src/keys.rs
```diff
@@ -39,18 +46,32 @@ pub trait SecretKey:
...
+#[cfg(all(feature = "serde", not(feature = "zero")))]
+pub trait SuperPublicKey:
+    ByteArray + Add<Output = Self> + Clone + PartialOrd + Ord + Default + Serialize + DeserializeOwned
+{
+}
+#[cfg(all(not(feature = "serde"), feature = "zero"))]
+pub trait SuperPublicKey: ByteArray + Add<Output = Self> + Clone + PartialOrd + Ord + Default + Zeroize {}
+#[cfg(all(not(feature = "serde"), not(feature = "zero")))]
+pub trait SuperPublicKey: ByteArray + Add<Output = Self> + Clone + PartialOrd + Ord + Default {}
+impl<T: PublicKey> SuperPublicKey for T {}
```
AaronFeickert, line 70: This naming also seems confusing.

AaronFeickert (changes requested):

SWvheerden:
this has been replaced by #191

PR #187, opened 2023-07-20
chore: update dependencies

hansieodendaal:
- Updated "tari_bulletproofs_plus".
- Updated "tari_utilities".

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ version = "0.17.0"
 edition = "2018"
 
 [dependencies]
-tari_utilities = { package = "tari_utilities", version = "0.5.1", features = ["zero", "std"] }
+tari_utilities = { version = "0.5.1", features = ["zero", "std"] }
 blake2 = { version = "0.10" }
 borsh = { version = "0.10" , optional = true }
 bulletproofs_plus = { package = "tari_bulletproofs_plus", version = "0.3.2" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,10 +11,10 @@ version = "0.17.0"
 edition = "2018"
 
 [dependencies]
-tari_utilities = { version = "0.5", features = ["zero", "std"] }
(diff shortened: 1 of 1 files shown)


AaronFeickert:
There are several (non-breaking) Bulletproofs+ fixes that aren't included in this PR's dependency revision. Might be worth waiting until a new version is tagged and released; then it can be referenced by crate version.

hansieodendaal:
> There are several (non-breaking) Bulletproofs+ fixes that aren't included in this PR's dependency revision. Might be worth waiting until a new version is tagged and released; then it can be referenced by crate version.

Absolutely, this can be referenced by commit until all is stable.

AaronFeickert:
There's still [one pending PR](https://github.com/tari-project/bulletproofs-plus/pull/50) that hasn't been merged, but it might be worthwhile to update to the [latest commit](https://github.com/tari-project/bulletproofs-plus/commit/b253a6fa4166e821a2103b2c4415bad2dfabf86d) for CI purposes.

AaronFeickert:
Crazy timing, but the pending PR was just merged, so there's a [newer commit](https://github.com/AaronFeickert/bulletproofs-plus/commit/b4756b05edfc2cc0f9b0240ca34a22bd9730fbd1) now.

AaronFeickert:
I assume a new BP+ version will be issued after any audit findings are addressed.

Cargo.toml
```diff
@@ -11,10 +11,10 @@ version = "0.17.0"
 edition = "2018"
 
 [dependencies]
-tari_utilities = { version = "0.5", features = ["zero", "std"] }
+tari_utilities = { package = "tari_utilities", version = "0.5.1", features = ["zero", "std"] }
```
AaronFeickert, line 14: Do you need the `package` specification here?

AaronFeickert (changes requested):

Cargo.toml
hansieodendaal: Not necessary, will remove

SWvheerden (approved):

PR #194, opened 2023-08-09
feat!: differentiate scalar parsing from byte arrays

### LINKED ISSUE #189: Differentiate scalar creation by canonical and reduction cases
Currently, instantiation of a `RistrettoSecretKey` using its `ByteArray` implementation does not differentiate between arrays that are expected to be canonical representations and those that are expected to use reduction.

These cases should be differentiated. The `ByteArray` implementation for `RistrettoSecretKey` should be modified to perform canonical byte parsing. A separate wide reduction method should be used for cases where a non-canonical representation is acceptable, and where wide reduction applies.

AaronFeickert:
Currently, creating a scalar `RistrettoSecretKey` [from a byte array](https://github.com/tari-project/tari-crypto/blob/053119f2110aaf3089c7b9df96f50b8cc8d3217a/src/ristretto/ristretto_keys.rs#L90-L100) performs modular reduction on 32 bytes. For cases where the input is intended to be canonical, this is suboptimal. For cases where the input is produced from a hashing operation, wide reduction should be used to mitigate bias.

This work renames `SecretKey::from_bytes` to `SecretKey::from_canonical_bytes` to support an underlying `ByteArray` trait update. In the case of `RistrettoSecretKey`, it uses the curve library's canonical parser and returns an error if the provided byte slice is not a canonical scalar encoding.

It also adds a new `SecretKey::from_uniform_bytes` function that uses wide reduction. For constructions like signatures and KDFs that use hashing operations to produce scalar values, this function is used and the underlying hashers are updated to produce 64-byte output in the case of `RistrettoSecretKey`.

It updates the Schnorr signature API to support raw signing and verification using challenge byte slices that are either canonical encodings or uniform. It renames several existing functions for clarity.

It corrects a few typos that were discovered along the way.

Closes #189.

BREAKING CHANGE: This changes the way that scalars are produced from byte arrays, modifies the `SecretKey` trait and corresponding `RistrettoSecretKey` implementation, and updates the Schnorr signature API.


AaronFeickert:
Marking as draft since this should not be merged until it can be tested against the `tari` repository.

src/ristretto/ristretto_keys.rs
```diff
@@ -74,29 +74,51 @@ impl borsh::BorshDeserialize for RistrettoSecretKey {
...
     }
+
+    /// Return a secret key computed by wide reduction on a byte array
+    /// If the byte array is not exactly 64 bytes, returns an error
+    fn from_bytes_wide(bytes: &[u8]) -> Result<Self, ByteArrayError> {
+        if bytes.len() != Self::WIDE_REDUCTION_LEN {
+            return Err(ByteArrayError::IncorrectLength {});
+        }
+
+        // We need a slice of the correct length, which can't be enforced by the function signature
```
CjS77, line 91: I found this comment confusing. 

You have to copy because `Scalar::from_bytes_mod_order_wide` accepts a reference to an array of 64 bytes, whereas `bytes` is a reference to a byte slice.

src/ristretto/ristretto_keys.rs
```diff
@@ -74,29 +74,51 @@ impl borsh::BorshDeserialize for RistrettoSecretKey {
...
+        if bytes.len() != Self::KEY_LEN {
             return Err(ByteArrayError::IncorrectLength {});
         }
-        let mut a = [0u8; 32];
-        a.copy_from_slice(bytes);
-        let k = Scalar::from_bytes_mod_order(a);
-        a.zeroize();
-        Ok(RistrettoSecretKey(k))
+
+        // We need a slice of the correct length, which can't be enforced by the function signature
```
CjS77, line 111:
```suggestion
        // We need an array of the correct length, which can't be enforced by the function signature
```

.. and anyway `Scalar::from_canonical_bytes` takes ownership

CjS77 (commented):
Some nits re: one of the comments. Otherwise LGTM

CjS77:
ACK

src/ristretto/ristretto_keys.rs
AaronFeickert: Right. The comment was intended to indicate that we can't have the function signature accept `bytes` as `&[u8; Self::WIDE_REDUCTION_LEN]` due to compiler restrictions, so the only solution is to do the copy. It wasn't explained well, so I just removed it.

src/ristretto/ristretto_keys.rs
AaronFeickert: Yep. I copied the other comment, which was inaccurate in this case. Removed.

hansieodendaal (commented):
Nice, just some comments:
- Why forcing `from_bytes` to use the underlying `from_canonical_bytes` call? Why not add an additional method `from_canonical_bytes` method to use the underlying `from_canonical_bytes` call? This way the change will be non-breaking.
- If the intention is to prohibit the underlying `from_bytes_mod_order` call, I would propose changing the name of `from_bytes` to  `from_canonical_bytes`.

src/hashing.rs
```diff
@@ -32,16 +32,29 @@ use alloc::string::String;
...
     alloc::string::ToString,
     errors::{HashingError, SliceError},
     keys::SecretKey,
 };
 
+/// Wrapper type for 256-bit `Blake2` hasher
+pub type Blake2b256 = Blake2b<U32>;
+
+/// Wrapper type for 512-bit `Blake2` hasher
+pub type Blake2b512 = Blake2b<U64>;
```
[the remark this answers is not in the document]
AaronFeickert: We certainly could, but then it requires us to include `use` declarations from both `blake2` and `digest` everywhere, which gets old pretty quickly.

AaronFeickert:
@hansieodendaal thanks for the review.

> * Why force `from_bytes` to use the underlying `from_canonical_bytes` call? Why not add an additional method `from_canonical_bytes` method to use the underlying `from_canonical_bytes` call? This way the change will be non-breaking.

There's no good reason to have scalar parsing use the underlying `Scalar::from_bytes_mod_order` call. I'd much rather have it be breaking and use a consistent approach.

> * If the intention is to prohibit the underlying `from_bytes_mod_order` call, I would propose changing the name of `from_bytes` to  `from_canonical_bytes`.

Unfortunately implementing `ByteArray` requires this.

src/hashing.rs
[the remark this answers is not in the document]
hansieodendaal: In `tari-project` we have done away with type aliases for `Blake2b<U32>` and `Blake2b<U64>` altogether, which I support.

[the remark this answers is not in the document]
AaronFeickert: OK, sounds good. I'll remove the aliases here.

hansieodendaal (approved):
ACK

src/ristretto/ristretto_keys.rs
```diff
@@ -74,29 +74,49 @@ impl borsh::BorshDeserialize for RistrettoSecretKey {
...
 //-------------------------------------  Ristretto Secret Key ByteArray  ---------------------------------------------//
 
 impl ByteArray for RistrettoSecretKey {
-    /// Create a secret key on the Ristretto255 curve using the given little-endian byte array. If the byte array is
-    /// not exactly 32 bytes long, `from_bytes` returns an error. This function is guaranteed to return a valid key
-    /// in the group since it performs a mod _l_ on the input.
+    /// Return a secret key computed from a canonical byte array
+    /// If the byte array is not exactly 32 bytes, returns an error
+    /// If the byte array does not represent a canonical encoding, returns an error
     fn from_bytes(bytes: &[u8]) -> Result<RistrettoSecretKey, ByteArrayError>
```
stringhandler, line 104: I would prefer to not implement ByteArray and instead have a specific `from_bytes_cannonical` or `from_bytes_exact` trait

stringhandler (changes requested):

src/ristretto/ristretto_keys.rs
AaronFeickert: Should other structs that have canonical representations (like Ristretto group elements) move to that trait as well?

stringhandler: AFAIK the ByteArray trait makes it simple to use `to_hex()` but doesn't really have any other significant addition, apart from a marker. If I were doing the change I would probably just add the methods directly onto the structs without a trait. Beware that there are also Serialize and BorshSerialize methods which should all use the canonical versions and error if not canonical

AaronFeickert:
After discussion, it sounds like the preferred approach is to rename the `ByteArray` function `from_bytes` to `from_bytes_canonical` and to have `SecretKey` use `from_bytes_wide`. This will reduce confusion and force all existing `from_bytes` uses to pick one or the other.

AaronFeickert:
This is being marked as ready to review. However, it must NOT BE MERGED until the `tari` [corresponding PR](https://github.com/tari-project/tari/pull/5636) has been reviewed and approved, in order to ensure that no further changes are necessary here.

hansieodendaal (approved):
Nice, sensible and easy to reason about.

ACK

Cargo.toml
```diff
@@ -11,7 +11,7 @@ version = "0.18.0"
 edition = "2018"
 
 [dependencies]
-tari_utilities = { version = "0.5", default-features = false, features = ["zero"] }
+tari_utilities = { git = "https://github.com/AaronFeickert/tari_utilities", branch = "byte-array", default-features = false, features = ["zero"] }
```
stringhandler, line 14: Needs update

src/ristretto/ristretto_keys.rs
```diff
@@ -66,37 +66,57 @@ impl borsh::BorshDeserialize for RistrettoSecretKey {
...
+    const WIDE_REDUCTION_LEN: usize = 64;
 
     /// Return a random secret key on the `ristretto255` curve using the supplied CSPRNG.
     fn random<R: RngCore + CryptoRng>(rng: &mut R) -> Self {
         RistrettoSecretKey(Scalar::random(rng))
     }
+
+    /// Return a secret key computed by wide reduction on a byte array
+    /// If the byte array is not exactly 64 bytes, returns an error
+    fn from_bytes_wide(bytes: &[u8]) -> Result<Self, ByteArrayError> {
```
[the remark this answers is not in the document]
AaronFeickert: How about `from_uniform_bytes`?

Cargo.toml
AaronFeickert: Done!

SWvheerden (commented):
LGTM

stringhandler (approved):

PR #202, opened 2023-08-27
tests: add fuzzing of secret key

stringhandler:
WIP: Added fuzzing to RistrettoSecretKey, this is difficult to test a round trip to and from bytes because the from_bytes implementation allows field values higher than the group order. I'm going to leave this in draft until we have merged the other PR that implements `from_canonical_bytes`

### DIFF
--- a/fuzz/.gitignore
+++ b/fuzz/.gitignore
@@ -1,0 +1,4 @@
+target
+corpus
+artifacts
+coverage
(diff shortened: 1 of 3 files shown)


stringhandler:
Note: you must update your rust-toolchain to a nightly build to use fuzzing
```toml
[toolchain]
#channel = "stable"
channel = "nightly-2023-06-04"

```

AaronFeickert:
The title prefix should be changed to `test` to stop CI from complaining.

PR #204, opened 2023-08-27
fix!: zeroize temporary scalar value

stringhandler:
Zeroize a temporary array of Scalars. Scalar is `Copy` so these would be placed into a new `Vec` in memory.

I have also removed `commit_scalars` from the public interface, since it is a foot gun for anyone who uses it but does not zeroize the blinding factors


AaronFeickert:
I don't think you need to do anything to `ExtendedMask` in order to zeroize on drop. The underlying `SecretKey` trait [must zeorize on drop](https://github.com/tari-project/tari-crypto/blob/1e6f60319d483bb504bead04c4f8e37a6734b770/src/keys.rs#L29-L30), and while the `zeroize` documentation notes some [caveats](https://docs.rs/zeroize/latest/zeroize/#stackheap-zeroing-notes) about heap reallocation, it does specifically indicate that `Vec` backing buffers are handled.

src/extended_range_proof.rs
```diff
@@ -212,6 +219,14 @@ where K: SecretKey
     pub minimum_value_promise: u64,
 }
 
+impl<K: SecretKey> Drop for ExtendedWitness<K>{
+    fn drop(&mut self) {
+        self.mask.zeroize();
+        self.value.zeroize();
+        self.minimum_value_promise.zeroize();
+    }
+}
```
AaronFeickert, line 228: You should be able to get this functionality for free by deriving `ZeroizeOnDrop`.

src/extended_range_proof.rs
```diff
@@ -99,13 +100,19 @@ pub trait ExtendedRangeProofService {
...
 where K: SecretKey
 {
     secrets: Vec<K>,
 }
 
+impl<K:SecretKey> Drop for ExtendedMask<K>{
+    fn drop(&mut self) {
+        self.secrets.zeroize();
+    }
+}
```
AaronFeickert, line 114: The `SecretKey` trait already enforces zeroizing on drop, so I don't think this is necessary.

AaronFeickert: If you still wanted to separately enforce zeroizing on drop at the `ExtendedMask` level as a safeguard against future structural changes, you should be able to simply derive `ZeroizeOnDrop` instead of this manual implementation.

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -166,8 +165,9 @@ impl ExtendedHomomorphicCommitmentFactory for ExtendedPedersenCommitmentFactory
         k_vec: &[RistrettoSecretKey],
         v: &RistrettoSecretKey,
     ) -> Result<PedersenCommitment, CommitmentError> {
-        let blinding_factors: Vec<Scalar> = k_vec.iter().map(|k| k.0).collect();
+        let mut blinding_factors: Vec<Scalar> = k_vec.iter().map(|k| k.0).collect();
```
AaronFeickert, line 168: Can you use a `Zeroizing` wrapper here instead? Then it'll drop if the `commit_scalars` call fails.

AaronFeickert (changes requested):

src/ristretto/pedersen/extended_commitment_factory.rs
stringhandler: good point, let me see

src/extended_range_proof.rs
AaronFeickert: As a side note, `ZeroizeOnDrop` also has the benefit of being future-proof against structural changes.

src/ristretto/pedersen/extended_commitment_factory.rs
AaronFeickert: It works! See [this branch](https://github.com/AaronFeickert/tari-crypto/commit/0a7e629c4cf4fa84af0957febbd9a62c1b4741f5) for an implementation.

src/extended_range_proof.rs
AaronFeickert: See [this branch](https://github.com/AaronFeickert/tari-crypto/commit/0a7e629c4cf4fa84af0957febbd9a62c1b4741f5) for the derived traits.

src/extended_range_proof.rs
AaronFeickert: See [this commit](https://github.com/AaronFeickert/tari-crypto/commit/83ee9b7f95d1491d4221d81fb4be33852499ab91) for the derived traits.

AaronFeickert (changes requested):

src/extended_range_proof.rs
AaronFeickert: Also, if you derive `ZeroizeOnDrop`, I'd also do `Zeroize` too; this will allow a caller to manually call `zeroize` if they really want to. You don't get `Zeroize` for free for `ZeroizeOnDrop`.

src/ristretto/ristretto_keys.rs
```diff
@@ -65,9 +65,19 @@ impl borsh::BorshSerialize for RistrettoSecretKey {
 impl borsh::BorshDeserialize for RistrettoSecretKey {
     fn deserialize_reader<R>(reader: &mut R) -> Result<Self, borsh::maybestd::io::Error>
     where R: borsh::maybestd::io::Read {
-        let bytes: Vec<u8> = borsh::BorshDeserialize::deserialize_reader(reader)?;
-        Self::from_bytes(bytes.as_slice())
+        let mut bytes: Vec<u8> = borsh::BorshDeserialize::deserialize_reader(reader)?;
```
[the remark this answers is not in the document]
stringhandler: I tried, but it doesn't deserialize, I could have added a wrapper after the deserialize, but it's basically the same thing

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -105,13 +104,13 @@ impl ExtendedPedersenCommitmentFactory {
             }
             #[cfg(not(feature = "precomputed_tables"))]
             {
-                let scalars = once(value).chain(blinding_factors);
+                let scalars = once(value).chain(blinding_factors.iter());
```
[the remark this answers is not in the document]
stringhandler: I think you are correct. We can leave it as is

src/ristretto/pedersen/extended_commitment_factory.rs
```diff
@@ -111,7 +110,7 @@ impl ExtendedPedersenCommitmentFactory {
                 Ok(RistrettoPoint::multiscalar_mul(scalars, points))
             }
         } else {
-            let scalars = once(value).chain(blinding_factors);
+            let scalars = once(value).chain(blinding_factors.iter());
```
stringhandler, line 113:
```suggestion
            let scalars = once(value).chain(blinding_factors);
```

src/ristretto/pedersen/extended_commitment_factory.rs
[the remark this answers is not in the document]
AaronFeickert: If my comment is correct, then I think you should leave the change in. That is, if (without `iter()`) the compiler is making a copy of `blinding_factors` and then providing references, the copy isn't zeroized and that's bad. If `iter()` ensures that a copy isn't made, all is good.

src/ristretto/ristretto_keys.rs
```diff
@@ -234,12 +244,6 @@ impl From<u64> for RistrettoSecretKey {
     }
 }
 
-impl From<Scalar> for RistrettoSecretKey {
-    fn from(s: Scalar) -> Self {
-        RistrettoSecretKey(s)
-    }
-}
-
```
AaronFeickert, line 242: This is a breaking change. Has it been confirmed that `tari` doesn't use this?

AaronFeickert: I just checked that this doesn't introduce problems against [this PR](https://github.com/tari-project/tari/pull/5636) in `tari` (checking against its main branch isn't possible), so removing the functionality is fine.

AaronFeickert: Probably worth updating the PR description to clarify that this is a breaking API change.

AaronFeickert:
Superseded by #209.

AaronFeickert:
Closing since #209 has been merged.

PR #205, opened 2023-08-27
ci: more jobs for precomputed_tables

stringhandler:
Since the features for precomputed_tables include different code when the feature is enabled or not, it is not sufficient to test with --all-features. In this PR I've tried to add jobs for running CI to find problems with code that is excluded when the feature is present or not


AaronFeickert (approved):

AaronFeickert:
This should be updated with the actions from #206 for consistency.

PR #217, opened 2023-12-18
fix: reject identity keys and commitments in signatures

AaronFeickert:
Signature verification currently allows identity elements:
- `SchnorrSignature` allows an identity public key
- `CommitmentSignature` allows an identity commitment
- `CommitmentAndPublicKeySignature` allows both

This doesn't strictly break soundness, but does remove message binding. While it shouldn't be problematic for unforgeability, it's a case that shouldn't arise from an honest signer and is easy to check for.

This PR fails signature verification if such an identity element is provided, and adds tests for each case. It does _not_ return an error if the corresponding identity elements are provided by the signer, since this precludes useful partial signature operations.

### DIFF
--- a/src/ristretto/ristretto_com_and_pub_sig.rs
+++ b/src/ristretto/ristretto_com_and_pub_sig.rs
@@ -82,10 +82,11 @@ pub type RistrettoComAndPubSig = CommitmentAndPublicKeySignature<RistrettoPublic
 mod test {
     use blake2::Blake2b;
     use digest::{consts::U64, Digest};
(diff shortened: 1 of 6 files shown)


AaronFeickert:
It should be manually checked that each test fails if the corresponding verification check is disabled, which confirms that each signature is otherwise valid.

AaronFeickert:
This currently fails CI due to a lint error that is fixed by #221.

AaronFeickert:
This fails CI and is blocked on the fix in #222.

PR #218, opened 2024-01-15
chore: update range proof dependency

AaronFeickert:
Updates the Bulletproofs+ range proof dependency. Makes a few updates to account for its API changes.


AaronFeickert:
Marked as draft until a new Bulletproofs+ library release is issued and this PR's `Cargo.toml` reference is updated correspondingly.

AaronFeickert:
Looks like something changed with nightly Clippy since the last merge that thinks there's recursion in the key methods. Unsure if the best approach is an `allow`, or a switch to stable or older nightly Clippy.

Would appreciate suggestions (ping @sdbondi @hansieodendaal @stringhandler).

hansieodendaal:
I think this change to stable clippy is good

AaronFeickert:
> I think this change to stable clippy is good

This is addressed directly in #221.

SWvheerden (approved):

PR #219, opened 2024-01-15
feat: add constant-time trait bounds

### LINKED ISSUE #139: Check that secret-type material uses constant-time comparisons
It's often important that secret types use constant-time comparisons to avoid leaking data. Much of this is abstracted by the use of upstream key types, but this should be checked for custom types.

For example, MACs and other derived keys almost certainly need this property. See the [subtle](https://docs.rs/subtle/latest/subtle/) crate for good information and tooling on this.

AaronFeickert:
Currently, the only implementation of the `SecretKey` and `PublicKey` traits is for Ristretto, where both [scalars](https://github.com/dalek-cryptography/curve25519-dalek/blob/ba737a379071191158bacfa6d138f6249b12fc09/curve25519-dalek/src/scalar.rs#L296-L300) and [group elements](https://github.com/dalek-cryptography/curve25519-dalek/blob/ba737a379071191158bacfa6d138f6249b12fc09/curve25519-dalek/src/ristretto.rs#L822-L826) use constant-time equality in their underlying `PartialEq` implementations, and which support the `ConstantTimeEq` trait.

This PR does what it can to encourage the use of constant-time equality for keys by doing a few things.

First, it requires that any types implementing `SecretKey` or `PublicKey` also implement `ConstantTimeEq`. Unfortunately, this doesn't guarantee that their `PartialEq` implementation defaults to this, and it doesn't appear possible to enforce this at the trait level.

It also sets a good example by manually implementing `PartialEq` on the Ristretto key types to use their `ConstantTimeEq` implementations. This isn't strictly necessary, but hopefully helps to indicate best practice. It also implements `ConstantTimeEq` directly as required by the new trait bounds.

Finally, it implements `ConstantTimeEq` for `DiffieHellmanSharedSecret` using the new trait bound, and removes a redundant `Zeroize` trait bound.

Note that this doesn't actually change the current implementations' behavior, and therefore incurs no performance hit.

Closes #139.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,6 +24,7 @@ rand_core = { version = "0.6" , default-features = false}
 serde = { version = "1.0", optional = true }
 sha3 = { version = "0.10", default-features = false  }
 snafu = { version = "0.7", default-features = false}
+subtle = { version = "2.5.0", default-features = false }
 zeroize = {version = "1" , default-features = false}
 
 [dev-dependencies]
@@ -48,6 +49,7 @@ std = [
     "serde?/std",
     "sha3/std",
     "snafu/std",
+    "subtle/std",
     "tari_utilities/std",
     "zeroize/std",
 ]
--- a/src/dhke.rs
+++ b/src/dhke.rs
@@ -11,18 +11,19 @@
 
 use core::ops::Mul;
 
+use subtle::{Choice, ConstantTimeEq};
 use zeroize::{Zeroize, ZeroizeOnDrop};
 
 use crate::keys::PublicKey;
 
 /// The result of a Diffie-Hellman key exchange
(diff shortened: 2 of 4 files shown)


AaronFeickert:
This currently fails CI due to a lint error that is fixed by #221.

sdbondi (approved):
utACK

sdbondi (approved):

SWvheerden (approved):

AaronFeickert:
This fails CI and is blocked on the fix in #222.

AaronFeickert:
I have an [open PR](https://github.com/dalek-cryptography/subtle/pull/131) in `subtle` to add a marker trait intended to signal that all equality testing for a type is done in constant time. If this is merged and then used in the `curve25519-dalek` types, it could be used as a trait bound here.

PR #223, opened 2024-02-07
ci: move to stable toolchain

AaronFeickert:
Currently, CI uses a mix of stable and nightly toolchains. Nightly is included primarily to support `no_std` checks, but this is only necessary due to a `rand` test dependency.

This PR makes changes to allow a move to stable across the board. To do this, it replaces the use of `rand` in all tests with a random number generator from `rand_chacha`, which has the side effect of making all tests deterministic.

However, due to a desire to keep a secure random number generator in doctests, we add an optional `rand` feature whose only role is to enable `rand_core::OsRng`. Doctests that use this are gated on this feature.

The `no_std` target has been changed to `wasm32-unknown-unknown` for easier local testing; it should still meet the goal.

Finally, note that some formatting rules are not available on stable, so we still use the latest nightly for this.

### DIFF
--- a/.github/workflows/clippy-check.yml
+++ b/.github/workflows/clippy-check.yml
@@ -1,9 +1,6 @@
 name: Formatting, lints, and code checks
 on: [push, pull_request]
 
-env:
-  nightly: nightly-2024-02-01
-
 jobs:
   clippy_check:
     runs-on: ubuntu-latest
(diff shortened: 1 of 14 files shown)


SWvheerden (commented):
I dont think we need to remove rand, rand should work on stable.

AaronFeickert:
Closing.

PR #224, opened 2024-02-24
feat: add range proof batch verification blame

AaronFeickert:
It may be the case that verification of a batch of range proofs fails. When this happens, the verifier does not know which proofs in the batch are invalid.

The best approach to take depends on the use case.

If you only need to know the index of _any_ invalid proof in the batch, a linear scan is inefficient. In this case, performing a binary search on the batch is more efficient since it requires only a logarithmic number of verifications.

On the other hand, if you need to know the index of _all_ invalid proofs in the batch, and if this number may be large, a complete linear search is more efficient than repeated binary searches.

This PR adds both approaches via the `ExtendedRangeProofService` trait. The new `verify_batch_with_first_blame` function performs a binary search if batch verification fails, returning an error containing the index of the first invalid proof. The new `verify_batch_with_all_blame` function performs a complete linear search if batch verification fails, returning an error containing the indexes of all failed proofs.

### DIFF
--- a/src/extended_range_proof.rs
+++ b/src/extended_range_proof.rs
@@ -83,6 +83,33 @@ pub trait ExtendedRangeProofService {
         statements: Vec<&AggregatedPublicStatement<Self::PK>>,
     ) -> Result<(), RangeProofError>;
 
+    /// Verify a batch of proofs.
+    /// If verification fails, this function performs a binary search to identify a failing proof and returns an error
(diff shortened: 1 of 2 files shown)


AaronFeickert:
This still needs tests.

AaronFeickert:
It was suggested that the Bulletproofs+ library would be a better place for this.

AaronFeickert:
Closing since this is likely better handled in the range proof library.

PR #226, opened 2024-04-12
fix: off by one error in domain seperation

bgorlick:
Looked like  there was a fixed length assumption for version bytes causing off-by-one errors with an inaccurate tag length calculation potentially leading to hash collisions. The static domain tag construction is not reflecting actual content variations.

Hopefully this addresses it:
We can dynamically construct the domain tags based on actual content, ensuring uniqueness. This should accurately calculate full the tag length to prevent inconsistencies, and then from there we can update the digest with exact length bytes and then the constructed tag. My thoughts is it's safer to converts version and label bytes to strings safely using String::from_utf8. this version implements expect for error handling to clearly identify conversion failures. I looked over the broader class and it looks like the approach is to panic, so I kept it consistent instead of returning String, which I had initially considered. 

Tested with build and cargo bench, and everything passes.

### DIFF
--- a/src/hashing.rs
+++ b/src/hashing.rs
@@ -78,26 +78,48 @@ pub trait DomainSeparation {
     /// Adds the domain separation tag to the given digest. The domain separation tag is defined as
     /// `{domain}.v{version}.{label}`, where the version and tag are typically hard-coded into the implementing
     /// type, and the label is provided per specific application of the domain.
+    ///
+    /// Revision with notes below:
+    /// Adds a domain separation tag to the digest.
+    /// The domain separation tag is constructed as `{domain}.v{version}.{label}`.
+    ///
+    /// # Arguments
+    /// * `digest` - A mutable reference to the digest to which the tag will be added.
+    /// * `label` - The label to use for creating the tag. If the label is empty, it will be omitted from the tag.
+    ///
+    /// # Panics
+    /// This function panics if the label or the version cannot be converted into a UTF-8 string. This can only occur if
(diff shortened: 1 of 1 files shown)


SWvheerden:
Thanks for the PR, 
Can you please run fmt by running:
```
Run cargo +nightly-2024-02-01 fmt --all -- --check
```

src/hashing.rs
```diff
@@ -78,26 +78,48 @@ pub trait DomainSeparation {
...
+
+        let version_string = String::from_utf8(version[version_offset..].to_vec())
+            .expect("Version conversion failed: invalid UTF-8");
+        let label_string = if label.is_empty() {
+            String::new()
         } else {
-            // 3 additional bytes are 2 x '.' delimiters and 'v' tag for version
-            domain.len() + (3 - version_offset) + label.len() + 3
+            String::from_utf8(label.to_vec())
+                .expect("Label conversion failed: invalid UTF-8")
```
SWvheerden, line 109: I am not a big fan of this approach. 
`S` is a str, now we convert it to bytes, only to convert it back to a string, then down below, convert the whole string back to bytes again?

bgorlick: You raise a valid point, I will share a variation. My main mo was to get something initially to address the issue. Will refactor.

AaronFeickert:
Can you provide an example of how the off-by-one error occurs? It's not immediately clear to me.

AaronFeickert:
Superseded by #229. Closing.

PR #229, opened 2024-05-09
feat!: update hasher design

AaronFeickert:
This PR updates the domain-separated hasher design.

Currently, the handling of domain separators is done via an initial length-prefixed human-readable concatenation of a domain separator, version number, and label that requires special handling due to encoding.

The design proposed here simplifies things. All input entering an underlying hash function (domain separator, version number, label, and any subsequent data) is prefixed by a tag byte that indicates its purpose. If the input is of variable length, it is prefixed by its length as well.

This removes the need for more complex encoding, is very robust against collisions, and is flexible in that it allows for other input types to be added in the future if desired.

Most of the required changes are to unit tests; some of this has been done already.


AaronFeickert:
This PR is incomplete, but I'm opening it as a draft to get feedback.

AaronFeickert:
This would supersede #226, which uses the existing domain separation design.

hansieodendaal (commented):
This seems robust, with a cleaner and simpler implementation. I am in favour.

src/hashing.rs
```diff
@@ -49,6 +48,26 @@ use crate::{
...
+#[repr(u8)]
+enum Flag {
+    /// An initial domain separator indicating the purpose of the hasher
+    DomainSeparator = 0,
+    /// The version of the hasher, which MUST be a single byte
+    Version,
+    /// A label that can be used to differentiate uses of the hasher
+    Label,
+    /// Arbitrary byte data to be added to the hasher
+    Data,
```
stringhandler, line 68: I wouldn't know when to use data and when to use label. Perhaps we only need one of them?

AaronFeickert: I'll update the documentation to make this distinction more clear.

AaronFeickert:
Blocked on a decision about whether to use a Borsh/CBOR design instead, as is used elsewhere.

src/hashing.rs
```diff
@@ -49,6 +48,28 @@ use crate::{
...
+///
+/// The idea is that when putting input into a hasher, we include several things:
+/// - A flag indicating the input type, encoded as a byte
+/// - If the input is of variable length (determined by the type), a little-endian 64-bit encoding of the length
+/// - The input, encoded as bytes
+/// By doing so, we mitigate the risk of collision.
+/// 
+/// These are not exposed to the user; they are used internally only.
+#[repr(u8)]
+enum Flag {
```
sdbondi, line 62: Maybe `Tag` or `HasherTag`?

sdbondi:
In favour of these simplifications and tagging. I don't think that it is a one or the other decision re DS borsh hashing and DS raw bytes hasher. They both have their place. I don't think we ever want to introduce CBOR into two crypto, that is firmly a concern of the L2

src/hashing.rs
AaronFeickert: Good call. I like `Tag` personally. Will update.

AaronFeickert:
> I would perhaps clarify that the update (and therefore chain) methods are not simply adding the given preimage bytes to the hasher, and these methods should be called in a consistent/canonical way for every given hash construction

Good call. I'll update the documentation accordingly.

hansieodendaal (approved):
ACK

PR #230, opened 2024-06-28
feat: add pub constructor to DHKE

SWvheerden:
add pub constructor to DHKE


brianp (approved):
I'm hesitant to say an actual constructor might be preferred, to still prevent direct access to the key. I think it's purpose is to try and prevent it leaking into memory where it's not zeroized
but i'm also okay with this for speed of release

brianp (approved):
I like the changes much more. Retains leak prevention of the inner secret, and allows us to create from an already known secret 👍🏻

PR #231, opened 2024-07-02
feat: add sum for schnorr signature and capk signature

SWvheerden:
Adds sum for schnorr signature and capk signature as a features
new version


brianp (approved):
LGTM

PR #232, opened 2024-07-03
feat: use constant-time equality checking for DHKE

AaronFeickert:
This PR ensures that `DiffieHellmanSharedSecret` equality testing is done in constant time.

Previously, this equality testing was offloaded to the underlying `PublicKey` type. While this type supports the `ConstantTimeEq` trait, it is not guaranteed that equality testing will use this in all implementations.

### DIFF
--- a/src/dhke.rs
+++ b/src/dhke.rs
@@ -18,7 +18,7 @@ use zeroize::{Zeroize, ZeroizeOnDrop};
 use crate::keys::PublicKey;
 
 /// The result of a Diffie-Hellman key exchange
-#[derive(PartialEq, Eq, Zeroize, ZeroizeOnDrop)]
+#[derive(Zeroize, ZeroizeOnDrop)]
 pub struct DiffieHellmanSharedSecret<P>(P)
 where P: PublicKey;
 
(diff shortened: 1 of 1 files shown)


stringhandler (commented):
While the intention to use this type is good, I think this type mostly is unused. It adds very little apart from being a marker trait. I feel that now that we are adding code to make it less vulnerable means that it is more harm than good. I would suggest that we instead delete this type and it's usages and replace them with the standard public key

stringhandler (commented):
NACK

stringhandler:
sorry, this PR is fine, but I think the type should be deleted instead

AaronFeickert:
@stringhandler I get where you're coming from, but the intent isn't for this to simple be a `PublicKey` wrapper. The result of a DHKE is a (shared) secret value, and the type makes it more difficult to misuse: it's zeroized on drop (and can be zeroized manually), intentionally doesn't let you do scalar or group arithmetic, and only supports canonical byte slice visibility.

Removing the type and simply using a `PublicKey` means you lose these protections and need to rely on the implementer, which seems unnecessary.

As to this PR specifically, it was noted in #219 that while we can require constant-time equality _support_ for `PublicKey` implementations, we can't enforce that all equality testing uses it. The addition of extra protections like forced constant-time equality in the DHKE type is meant to make up for this.

stringhandler (approved):
After discussing this offline, I'm happy to merge it

AaronFeickert:
Blocked on an unsigned commit.

PR #245, opened 2025-04-02
fix: check the key byte length generically

sdbondi:
CompressedKey<T> assumes the key length is 32 in `from_canonical_bytes`.

Made `CompressedKey::new` private since it is not used externally and needs to be used correctly, as it does not check byte length or if the compressed key bytes are valid

## Summary by CodeRabbit

- **Refactor**
  - Enhanced key management by refining validation routines to enforce dynamic key length requirements.
  - Improved encapsulation of internal key creation, ensuring that constraints are consistently applied for improved reliability.

### DIFF
--- a/src/compressed_key.rs
+++ b/src/compressed_key.rs
@@ -78,8 +78,8 @@ impl<T: PublicKey> CompressedKey<T> {
 }
 
 impl<T> CompressedKey<T> {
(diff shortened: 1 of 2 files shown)


sdbondi:
Looks like more work is needed if ByteArray impl has the PublicKey trait bound - closing for now

PR #247, opened 2025-07-23
fix: adds secret key conversions

sdbondi:
Adds the same conversions to RistrettoSecretKey supported by dalek's Scalar.
Adds `RistrettoSecretKey::ZERO` constant

### DIFF
--- a/src/ristretto/ristretto_keys.rs
+++ b/src/ristretto/ristretto_keys.rs
@@ -52,8 +52,13 @@ use crate::{
 /// let _k2 = RistrettoSecretKey::from_hex(&"100000002000000030000000040000000");
 /// let _k3 = RistrettoSecretKey::random(&mut rng);
 /// ```
-#[derive(Clone, Default, Zeroize, ZeroizeOnDrop)]
+#[derive(Clone, Zeroize, ZeroizeOnDrop)]
 pub struct RistrettoSecretKey(pub(crate) Scalar);
+
+impl RistrettoSecretKey {
+    /// A 0 secret key
+    pub const ZERO: Self = RistrettoSecretKey(Scalar::ZERO);
+}
 
 #[cfg(feature = "borsh")]
 impl borsh::BorshSerialize for RistrettoSecretKey {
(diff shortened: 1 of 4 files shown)


src/ristretto/ristretto_keys.rs
```diff
@@ -236,10 +241,45 @@ define_mul_variants!(
 
 //---------------------------------------------      Conversions     -------------------------------------------------//
 
+impl From<u8> for RistrettoSecretKey {
```
sdbondi: _Posted by Claude._

Adding `From<u8|u16|u32|u128>` next to `From<u64>` makes `RistrettoSecretKey::from(<untyped literal>)` ambiguous. This PR already had to change `from(1024)` to `from(1024u64)` in its own tests for that reason.

Ootle calls it this way in several places, and they will stop compiling on `cargo update`:
- `crates/engine_types/src/crypto/elgamal.rs:314,328`
- `crates/engine_types/src/crypto/helpers.rs:125`
- `crates/template_test_tooling/src/support/stealth.rs:222`
- wallet `sdk_tests` / `storage_sqlite` tests

Cargo's semver guide treats inference breakage from new impls as a minor change, so this is technically fine in a patch release. It still breaks builds in practice, so either fix those call sites in Ootle before or alongside the upgrade, or drop the extra integer widths and keep only `From<u64>` + `From<Scalar>`.

sdbondi (commented):
_Posted by Claude._

The change looks good. `ZERO`, the manual `Default` and `From<Scalar>` are all fine. `From<Scalar>` doesn't widen the public dalek surface, since `From<RistrettoSecretKey> for Scalar` already exists.

One thing to consider before shipping this as a patch release: the new integer `From` impls break type inference for untyped literals downstream (see the inline comment). Separately, the PR title uses `fix:`, but this is a `feat:`.

src/ristretto/ristretto_keys.rs
sdbondi: True, I think this will cause an unexpected headache for people if it's a patch release, so we'll bundle it into the next minor release instead.

brianp (approved):
LGTM

PR #249, opened 2026-09-16
feat: vartime batch Schnorr verification

sdbondi:
## Description

Adds variable-time batch verification for Schnorr signatures.

Verifying `n` signatures one at a time costs `n` independent double-base multiplications. Rolling them into a single weighted equation

```text
Σ zᵢ·Rᵢ + Σ (zᵢ·eᵢ)·Pᵢ − (Σ zᵢ·sᵢ)·G == 0
```

turns that into one `2n+1`-term multiscalar multiplication. Every input is public data, so this can run in variable time, which is where most of the speed-up comes from.

Measured on ristretto255, this is **1.64x** faster at n = 2, **3.37x** at n = 16 and **5.84x** at n = 1024, and it does not lose at n = 1. Full table below.

New API:

- `SchnorrSignature::challenge_scalar` — exposes the `e = H(R, P, m)` that `verify` already derives internally, so callers building their own verification equations no longer have to reproduce the challenge derivation by hand.
- `SchnorrSignature::verify_batch` — Ristretto-only inherent method (generic over the signature's hash domain `H`). Weights derived from a domain separated hash of the whole batch. Every caller reaches the same verdict on the same bytes, so it is usable for consensus; soundness is the Fiat-Shamir argument.
- `SchnorrSignature::verify_batch_with_rng` — the same check with weights drawn from a caller-supplied CSPRNG, for callers that don't need reproducibility across parties.

The batch methods are an inherent impl on `SchnorrSignature<RistrettoPublicKey, RistrettoSecretKey, H>` rather than generic over `PublicKey`. That keeps the `PublicKey` trait untouched — no variable-time multiscalar method sitting next to `batch_mul` with an identical signature that a caller could feed a secret key to — and lets the implementation work in dalek's `Scalar` / `RistrettoPoint` directly. `challenge_scalar` stays generic, since it isn't variable-time and is useful on any curve.

### Design notes

- **Weights** are 128-bit with the top bit forced set, bounding the soundness error at `2^-127`. They stop a set of individually invalid signatures from cancelling one another out — see the `rejects_cancelling_errors` test.
- **The identity public key is rejected up front.** Under `P = 0` the batch equation degenerates to `s·G == R`, which anyone can satisfy, and the batch never falls through to the per-signature check that would refuse it. The check is on the decoded point, not on bytes.
- **Only a `bool` is returned.** Naming the offending index would need a double-base multiplication per term, which makes a rejected batch cost *more* than an accepted one — a free DoS wherever rejected work isn't paid for. Callers that want the culprit can loop `verify`.
- **An empty batch verifies vacuously** (documented on both entry points).
- **Variable time is safe here** because every input to a signature verification is public — the signature and public key travel on the wire, and the challenge derives from those plus the message. The one exception, documented on both entry points: if the *message* is confidential, the timing does depend on `H(R, P, m)`, and `verify` is the constant-time choice.
- **The generator is folded into the multiscalar multiplication** rather than multiplied separately against dalek's precomputed basepoint table. The table makes `k·G` fast in isolation, but it is still an entire extra scalar multiplication; at small `n` that fixed cost outweighs one more term in the MSM. Measured against a split form that does use the table: **−16.3%** at n = 1, −6.5% at n = 2, −7.2% at n = 4, −3.3% at n = 8, a wash from n = 16 to 256 (overlapping intervals), and **+7.8%** at n = 1024. The table is worth a flat ~13 µs on the generator term, so folding only pays while that constant is large relative to the multiplication it is folded into.

Folding is therefore the right default for this crate's first consumer — Ootle verifies one seal plus one authorization — while being neutral across most of the range and costing ~8% only at n = 1024. A size threshold would recover that one case, but the region below it is mostly ties, so it would buy a second code path for very little.

An earlier revision of this PR claimed folding won at every size by up to 40%. That comparison was wrong: its split arm used `RISTRETTO_BASEPOINT_POINT * scalar` (generic variable-base) rather than the precomputed table, handicapping it by ~13 µs at every size. The figures above are the corrected measurement.
- The deterministic weight transcript uses the labels `batch_weight` / `batch_weight_expand` under the signature's own `H` domain, so it cannot collide with the `challenge` label.
- `no_std` + `alloc` only; nothing new is pulled in.

## Motivation and Context

Ootle (`tari_ootle_transaction`) grew its own batch verifier, which forced it to depend on `curve25519-dalek` directly and to reproduce this crate's challenge derivation by hand. Both of those belong here. Ootle is the first consumer; other non-consensus callers (wallets, indexers) can use the RNG variant.

## How Has This Been Tested?

Unit tests (`src/ristretto/ristretto_sig.rs`, `src/signatures/schnorr.rs`):

- Both batch verifiers agree with per-signature `verify` for n in {1, 2, 3, 4, 8, 33}, with a foreign signature planted at every position for small batches and at the ends and middle for large ones. Negative cases assert that `verify` rejects too, so a passing negative test can't be one both verifiers happen to accept.
- Rejects cancelling errors (a `+δG` / `−δG` pair).
- Rejects permuted components (public key, public nonce, or signature scalar swapped across terms).
- Rejects the identity public key, alone and at either end of a batch with a valid signature.
- Mixed messages in one batch, including the same signatures re-paired with the wrong messages.
- Deterministic weights are stable across calls, distinct per item, and change when the message, public key, public nonce, signature scalar, batch length, or batch order changes.
- Weights are never zero, and encode the expected little-endian 128-bit integer.
- `challenge_scalar` agrees with `verify_challenge_scalar`, and a custom hash domain batches under its own domain.

Benchmark (`benches/signatures.rs`) covers per-signature vs. both batch variants for n in {1, 2, 4, …, 1024}, on ristretto255. Each size is signed once outside the measurement and every iteration verifies that same fixed input; re-signing per iteration swamps a difference of tens of microseconds in setup noise.

| n | individually | deterministic weights | speed-up | random weights | speed-up |
|---|---|---|---|---|---|
| 1 | 33.6 µs | 32.2 µs | **1.05x** | 33.0 µs | **1.02x** |
| 2 | 66.9 µs | 40.8 µs | **1.64x** | 41.5 µs | **1.61x** |
| 4 | 135.8 µs | 57.8 µs | **2.35x** | 58.2 µs | **2.33x** |
| 8 | 270.6 µs | 92.1 µs | **2.94x** | 92.4 µs | **2.93x** |
| 16 | 538.8 µs | 159.8 µs | **3.37x** | 159.1 µs | **3.39x** |
| 32 | 1.08 ms | 296.1 µs | **3.64x** | 295.2 µs | **3.65x** |
| 64 | 2.15 ms | 601.4 µs | **3.58x** | 567.2 µs | **3.80x** |
| 128 | 4.31 ms | 1.01 ms | **4.25x** | 1.01 ms | **4.26x** |
| 256 | 8.62 ms | 1.81 ms | **4.76x** | 1.79 ms | **4.83x** |
| 512 | 17.21 ms | 3.29 ms | **5.23x** | 3.20 ms | **5.38x** |
| 1024 | 34.40 ms | 5.89 ms | **5.84x** | 5.73 ms | **6.01x** |

Criterion medians, 10s measurement windows, single-threaded and pinned to one physical core with its SMT sibling idle, on an otherwise quiet machine. Confidence intervals were within about 1% of the median throughout. Note this box scales between 582 MHz and 5086 MHz under the `powersave` governor, so absolute figures are machine- and clock-state-dependent; the speed-up ratios are measured within a single run and are the meaningful number.

The two weight modes cost the same within noise. Deriving deterministic weights is one Blake2b-512 pass over the batch plus one digest per four items — about 0.3 µs at n = 1, which is why the deterministic variant is not measurably behind the random one even at the smallest size.

**n = 1 is the honest edge case.** It is break-even to modestly ahead: this table says 1.05x, while a separate run on a different random batch put it at 1.20x. With only three group elements in the multiplication, the variable-time cost depends noticeably on the particular scalars, so treat n = 1 as "does not lose" rather than as a specific speed-up. Every size from n = 2 up is a clear win.

`cargo test --all-features`, `cargo lints clippy --all-targets --all-features`, `cargo +nightly fmt --check` and `cargo check --no-default-features` are all clean.

## Breaking Changes

- [x] None
- [ ] Requires data directory on base node to be deleted
- [ ] Requires hard fork
- [ ] Other - Please specify

The `PublicKey` trait is untouched, and the new methods are inherent on the Ristretto signature type, so nothing existing changes shape.


sdbondi (commented):
Reviewed at 1898f50. No blocking findings.

**Verified**

- The batch equation in `verify_batch_with_weights` (`src/signatures/schnorr.rs:536`) is the correct weighted form of `s·G == R + e·P`, and `challenge_scalar` reuses `construct_domain_separated_challenge` with `Blake2b<U64>` and wide reduction, so `e` is bit-for-bit what `verify` derives.
- Identity-key rejection mirrors the existing check in `verify_challenge_scalar` (`src/signatures/schnorr.rs:239`), so batch and per-signature verdicts stay consistent on that input.
- The claim that `DomainSeparatedHasher::update` length-prefixes each field holds (`src/hashing.rs:276`), so the deterministic transcript is unambiguous across variable-length messages.
- Weight encoding: top bit forced, low 16 bytes of a zeroed `WIDE_REDUCTION_LEN` buffer, so the weight is the 128-bit integer with 127 bits of entropy. Soundness bound of `2^-127` is right for both variants.
- Ristretto `vartime_batch_mul` mirrors `batch_mul` one-for-one, and the trait default keeps out-of-tree implementors compiling.
- Locally: `cargo test --all-features` (134 pass), `cargo lints clippy --all-targets --all-features`, `cargo +nightly fmt --check`, `cargo check --no-default-features` all clean. The remaining clippy warnings are the pre-existing `non_local_definitions` lints from `Zeroize` derives, not from this PR.
- CI is green on all five checks.

**Non-blocking**

- `verify_batch_with_weights` (`src/signatures/schnorr.rs:566`) does a separate constant-time basepoint multiplication for `(Σ zᵢ·sᵢ)·G`. Folding `-(Σ zᵢ·sᵢ)` and `G` into the same `vartime_batch_mul` call and comparing against the identity would make it a single MSM. Whether that beats the precomputed basepoint table is worth one bench run, not a change on faith.
- The repo's recent history bumps the version in separate `chore: new release` commits. Including the bump here is fine if that's the intent, just noting it differs from the last three releases.

**Nits**

- `agrees_with_per_signature_verify` (`src/ristretto/ristretto_sig.rs:235`) asserts `verify_batch` rejects directly and then again via `assert_rejects` on the next line; the first assert is redundant.
- `deterministic_weights_change_with_the_batch` (`src/signatures/schnorr.rs:723`) hand-copies `Signed` twice; deriving `Clone` on the test struct would remove both map blocks.

sdbondi (commented):
Re-reviewed at ea693be (diff from 1898f50).

- Both test nits addressed: the redundant `verify_batch` assert in `agrees_with_per_signature_verify` is gone, and `Signed` in `src/signatures/schnorr.rs` now derives `Clone`, replacing both hand-copy blocks. The added `case` labels on `assert_agrees`/`assert_accepts`/`assert_rejects` make a failing negative case identifiable, which the old `assert!(!assert_agrees(..))` did not.
- Version bump reverted to `0.23.3`, matching the repo's separate release-commit convention. PR body no longer claims the bump.
- No non-test code changed; the MSM fold suggestion remains optional.
- Verified in a detached worktree at ea693be: `cargo test --all-features` (134 pass), `cargo lints clippy --all-targets --all-features` (only pre-existing `non_local_definitions` lints), `cargo +nightly fmt --check` clean. CI green on all five checks.

Nothing outstanding from my side. GitHub blocks approval from the author's account, so this needs a second reviewer to land the approval.

sdbondi (commented):
Non-blocking suggestion: `deterministic_batch_weights` could return an iterator instead of a `Vec`.

The expansion is a pure function of `(seed, block)`, and `seed` is a fixed 64-byte digest, so after the transcript pass nothing needs to borrow `items`:

```rust
fn deterministic_batch_weights<B: AsRef<[u8]>>(
    items: &[(&Self, &P, B)],
) -> impl Iterator<Item = K> {
    let mut transcript = DomainSeparatedHasher::<Blake2b<U64>, H>::new_with_label("batch_weight");
    transcript.update((items.len() as u64).to_le_bytes());
    for (signature, public_key, message) in items {
        transcript.update(message.as_ref());
        transcript.update(public_key.as_bytes());
        transcript.update(signature.public_nonce.as_bytes());
        transcript.update(signature.signature.as_bytes());
    }
    let seed: [u8; 64] = transcript.finalize().as_ref().try_into().expect("Blake2b<U64> is 64 bytes");
    let n = items.len();

    (0..n.div_ceil(WEIGHTS_PER_DIGEST))
        .flat_map(move |block| {
            let digest = DomainSeparatedHasher::<Blake2b<U64>, H>::new_with_label("batch_weight_expand")
                .chain(seed)
                .chain((block as u64).to_le_bytes())
                .finalize();
            let (chunks, _) = digest.as_ref().as_chunks::<BATCH_WEIGHT_LEN>();
            let arr: [[u8; BATCH_WEIGHT_LEN]; WEIGHTS_PER_DIGEST] = chunks.try_into().expect("4 chunks");
            arr.map(Self::batch_weight_from_bytes) // [K; 4], IntoIterator by value
        })
        .take(n)
}
```

`verify_batch_with_weights` would then take `impl IntoIterator<Item = K>`, and `verify_batch_with_rng` can pass its `(0..n).map(..)` straight through, so both paths drop the weights `Vec`.

Caveats:

- `take` over `flat_map` is not `ExactSizeIterator`, so the `debug_assert_eq!(items.len(), weights.len())` goes away and `zip` will silently truncate on a length mismatch. Worth keeping a length check somewhere, or wrapping in a small counting struct.
- It removes one allocation of 32n bytes. The `scalars`/`points` vectors are still built, and dalek's `vartime_multiscalar_mul` collects its inputs and builds NAF and lookup tables internally, so the MSM dominates allocations. Expect the bench to move within noise.
- Mild real win: the block digests are computed lazily, so an early rejection (identity key at index 0) never hashes the remaining blocks.

Tests only need a `.collect::<Vec<_>>()` in the `weights` helper.

sdbondi (commented):
Re-reviewed at f50f154 (diff from ea693be). The refactor is sound; two doc-level leftovers below, nothing blocking.

**Verified**

- `PublicKey` and `RistrettoPublicKey` are back to `main` byte-for-byte, so the trait surface is unchanged and the "no vartime method next to `batch_mul`" rationale in the PR body holds.
- The folded equation `Σ zᵢ·Rᵢ + Σ (zᵢ·eᵢ)·Pᵢ − (Σ zᵢ·sᵢ)·G == 0` is checked with `is_identity()` on the MSM output (`src/ristretto/ristretto_sig.rs:264`). `-signature_sum` is pushed as the last scalar with `RISTRETTO_BASEPOINT_POINT` as its point, lengths stay equal, and `IsIdentity` is imported for `RistrettoPoint`.
- Identity rejection now checks `public_key.point().is_identity()` on the decoded point rather than a `ct_eq` against `P::default()`; same verdict on Ristretto since the encoding is canonical.
- `batch_weight_from_bytes` collapsed to `Scalar::from(u128 | 1 << 127)`, which is the same 128-bit little-endian value as the old wide-reduction path, and `weight_encoding` still pins it. The `Option` plumbing is gone with it, correctly, since nothing can fail any more.
- Weight/transcript tests moved into `ristretto_sig::test::batch` and still cover stability, per-field sensitivity, length, and order.
- Locally at f50f154: `cargo test --all-features` (134 pass), `cargo lints clippy --all-targets --all-features` (only pre-existing `non_local_definitions`), `cargo +nightly fmt --check`, `cargo check --no-default-features` all clean.

**Non-blocking**

- `src/ristretto/ristretto_sig.rs:219-224`: the doc comment on `verify_batch_with_weights` is two merged blocks. The first paragraph ("Reduces a batch to the scalars and points ... less the generator term. Returns `None` if any term is unusable ...") describes the `batch_terms` helper from an intermediate version; the function returns `bool` and does include the generator term. Drop the first block and keep the "Checks the batch equation ..." one.
- PR body, last line under Breaking Changes: "`PublicKey::vartime_batch_mul` is a new trait method with a default implementation" is stale now that the trait is untouched. The "New API" section above it is already correct.

sdbondi (commented):
Re-reviewed at 1275bb7 (diff from f50f154). Nothing blocking.

**Verified**

- Stale merged doc block on `verify_batch_with_weights` is gone; the stale `PublicKey::vartime_batch_mul` line is gone from the PR body.
- Bench change is sound: `batch` and `items` are built once per `n` outside the timed closure, all three arms verify the same fixed input, and `BatchSize` is still used by the sign/verify benches so the import stays. Verification is stateless, so reuse only lowers variance as the comment says.
- Locally at 1275bb7: `cargo test --all-features` (134 pass), `cargo lints clippy --all-targets --all-features` (benches included; only pre-existing `non_local_definitions`), `cargo +nightly fmt --check`, `cargo check --no-default-features` all clean. CI green on all five checks.

**One thing to confirm**

The doc on `verify_batch_with_weights` (`src/ristretto/ristretto_sig.rs:222`) and the PR body both say folding was measured against "dalek's precomputed basepoint table". The split arm I saw in the working tree during the earlier `verify_batch_with_weights_folded` experiment computed `RISTRETTO_BASEPOINT_POINT * signature_sum`, which is the constant-time variable-base multiplication, not `&RISTRETTO_BASEPOINT_TABLE * &scalar`. The two differ by roughly 3–4x for the generator term, which is the entire fixed cost the comparison hinges on. If the corrected figures came from a split arm using the table, the claim and the "no threshold warranted" conclusion stand. If not, the split numbers are handicapped and the comparison should be redone with the table before the doc asserts it. Either way, a one-line note in the doc naming which basepoint path was measured would make the figure reproducible.

sdbondi (commented):
Re-reviewed at c183915 (diff from 1275bb7). Doc-only change; nothing outstanding.

- The `verify_batch_with_weights` doc (`src/ristretto/ristretto_sig.rs:222-228`) now states the split arm used the basepoint table and gives the corrected figures, including the ~8% loss at n = 1024. That resolves the reproducibility point from my last review.
- The PR body matches the doc and records why the earlier figures were wrong. The "no threshold" reasoning is now backed by the measurement: the region below n = 1024 is ties, so a second code path would buy one size.
- No code changed. Locally at c183915: `cargo test --all-features` (134 pass), `cargo +nightly fmt --check` clean; clippy and no_std were clean at the previous commit and nothing outside a doc comment moved.

Nothing left from my side once CI finishes. Approval still needs a second reviewer since the author's account cannot approve.

src/signatures/schnorr.rs
```diff
@@ -243,6 +243,20 @@ where
...
     }
 
+    /// Returns the challenge scalar `e = H(R, P, m)` that [`SchnorrSignature::verify`] checks this signature
+    /// against.
+    ///
+    /// This is exactly the derivation used by [`SchnorrSignature::sign`] and [`SchnorrSignature::verify`]: the
+    /// domain separated Blake2b-512 challenge of [`SchnorrSignature::construct_domain_separated_challenge`],
+    /// wide-reduced into a scalar. It is exposed so that callers building their own verification equations do not
+    /// have to reproduce it.
+    pub fn challenge_scalar<B>(&self, public_key: &P, message: B) -> Result<K, SchnorrSignatureError>
```
SWvheerden, line 254: `verify` just below still carries its own copy of this derivation, so the thing `challenge_scalar` exists to stop callers doing — reproducing it by hand — is still being done once inside the crate. Two copies can drift, and only `challenge_scalar_matches_verify` would catch it.

Delegating removes the duplication structurally. I tried it and all 118 tests pass:

```rust
pub fn verify<'a, B>(&self, public_key: &'a P, message: B) -> bool
where
    for<'b> &'b K: Mul<&'a P, Output = P>,
    for<'b> &'b P: Add<P, Output = P>,
    B: AsRef<[u8]>,
{
    let Ok(e) = self.challenge_scalar(public_key, message) else {
        return false;
    };
    self.verify_challenge_scalar(public_key, &e)
}
```

src/ristretto/ristretto_sig.rs
```diff
@@ -124,6 +148,161 @@ pub type CompressedRistrettoSchnorr =
...
+    /// them and the message — so there is nothing secret for the timing to leak. If you are verifying a signature
+    /// over a message you are treating as confidential, the timing does depend on `H(R, P, m)`; use the
+    /// constant-time [`SchnorrSignature::verify`] for that.
+    ///
+    /// # Return value
+    ///
+    /// Only a `bool` is returned. Naming the offending index would need a double-base multiplication per term,
+    /// making a rejected batch cost more than an accepted one; callers that need the culprit should fall back to
+    /// [`SchnorrSignature::verify`] in a loop.
+    pub fn verify_batch<B: AsRef<[u8]>>(items: &[(&Self, &RistrettoPublicKey, B)]) -> bool {
```
SWvheerden, line 189: Minor ergonomics: `&[(&Self, &RistrettoPublicKey, B)]` means a caller holding owned data has to materialise an intermediate `Vec` of tuples-of-references before it can call this. The benchmark in this PR does exactly that:

```rust
let items: Vec<_> = batch.iter().map(|(s, p, m)| (s, p, m.as_slice())).collect();
```

which is the tell that every consumer will hit it. An `impl IntoIterator<Item = (&Self, &RistrettoPublicKey, B)>` or an owned-tuple form would avoid the copy. Not worth blocking on, and changing it later is not breaking for callers passing a slice, so this is fine to defer.

src/ristretto/ristretto_sig.rs
```diff
@@ -124,6 +148,161 @@ pub type CompressedRistrettoSchnorr =
...
+        if items.is_empty() {
+            return true;
```
SWvheerden, line 191: The empty case is documented on both entry points and matches the `Iterator::all` convention, so I wouldn't change the API.

But this is a consensus-facing entry point, and "zero signatures therefore valid" is the classic vacuous-truth footgun — it fails open, and it fails open silently. Worth confirming on the Ootle side that `verify_batch` can't be reached with an empty slice: if a transaction with no authorizations is representable at all, the non-empty check belongs at that call site rather than being left implicit here.

src/ristretto/ristretto_sig.rs
```diff
@@ -124,6 +148,161 @@ pub type CompressedRistrettoSchnorr =
...
+        }
+        let weights = Self::deterministic_batch_weights(items);
```
SWvheerden, line 197: Ordering nit, no action needed: the weight transcript hashes the whole batch here, before `verify_batch_with_weights` gets to the identity-key check, so a batch of all-identity keys pays a full Blake2b pass over every item before being rejected.

It's negligible next to the multiscalar multiplication it avoids — noting it only so the ordering is a deliberate choice rather than an accident, in case the rejection path ever gets scrutinised for DoS cost.

benches/signatures.rs
```diff
@@ -68,8 +68,55 @@ fn verify_message(c: &mut Criterion) {
...
+                }
+            });
+        });
+
+        group.bench_with_input(BenchmarkId::new("deterministic weights", n), &n, |b, _| {
+            b.iter(|| assert!(RistrettoSchnorr::verify_batch(&items)));
+        });
+
+        group.bench_with_input(BenchmarkId::new("random weights", n), &n, |b, _| {
+            b.iter(|| assert!(RistrettoSchnorr::verify_batch_with_rng(&items, &mut rng())));
```
SWvheerden: `rng()` is constructed inside the timed closure, so a `ThreadRng` handle is built on every iteration and counted in the measurement. It's a cheap TLS access and won't move the numbers meaningfully, but the deterministic arm has no equivalent per-iteration cost, so hoisting it makes the two arms strictly comparable:

```rust
let mut rng = rng();
group.bench_with_input(BenchmarkId::new("random weights", n), &n, |b, _| {
    b.iter(|| assert!(RistrettoSchnorr::verify_batch_with_rng(&items, &mut rng)));
});
```

Given the PR turns on differences of tens of microseconds at small `n`, it's worth removing the asymmetry.

SWvheerden (commented):
Reviewed the batch verifier. **The math and the security reasoning hold up**, and I reproduced the performance claims independently — including the fold-vs-split comparison, which came out the same shape as the corrected third revision (wins at small `n`, a wash mid-range, marginally behind at `n = 1024`). Good that that one got chased down rather than left standing.

Things I checked rather than read:

- `cargo test --all-features` (118 pass), `cargo clippy --all-targets --all-features`, and `cargo check --no-default-features` are all clean. The one `no_std` warning is pre-existing on `main`.
- **20k random batches fuzzed against per-signature `verify`: 0 mismatches** (2928 accepted / 17072 rejected), tampering across message, public key, nonce, scalar, identity key, zero signature, cross-term key swap, and untouched, at n in 1..6.
- Speed-ups reproduce: 1.11x at n=1, 1.73x at n=2, 3.40x at n=16, 5.62x at n=1024 (PR table: 1.05 / 1.64 / 3.37 / 5.84).

A few things I specifically tried to break and couldn't:

- The length-prefixing claim in the weight-transcript doc is real — `DomainSeparatedHasher::update` prepends a u64-LE length (`src/hashing.rs:276`), so the argument that a variable-length message can't be shifted across a field boundary holds.
- **Committing to `s` in the transcript is the load-bearing detail.** Without it an attacker could compute the weights and then solve for the last free variable; with it, the attack becomes a Fiat-Shamir fixed point, which is the claimed `2^127` search.
- Duplicate-defect cancellation is structurally impossible, not just improbable: n items sharing a defect need `sum(z_i) = 0 mod l`, and with weights in `[2^127, 2^128)` you'd need more than `2^124` terms to reach the group order.
- The identity-key rejection is the same predicate as `verify_challenge_scalar`'s zero-key check, so no input is accepted by one verifier and refused by the other.

The tests are unusually good. `assert_rejects` asserting that the *per-signature* verifier also rejects is the right instinct — it makes it impossible for a negative test to pass because both verifiers happen to accept, which is the exact failure mode that makes batch-verification suites worthless.

Comments below are all non-blocking. LGTM.

SWvheerden (approved):

